PR #1, opened 2015-05-25
Project structure

arturdryomov:
A very basic project structure to start with.

At this point the application can fetch tenants from a user-specified URL. This is done via AeroGear Pipe module for Android. I have a plan to consolidate all pipes-related managing and configuration at a `BackendClient` class to hide as much as possible from an actual client, i. e. UI. The client itself is delivered via dependency injection. At this point it seems a little bit silly — I can just use a singleton of `BackendClient` and be with it (a singleton mostly because it should save state between activities — pipes configuration shouldn’t be done very often and server address shouldn’t be passed on every client call), but probably it will help to deliver other object dependencies in the future.

There are some neat tricks which should help in the future as well — the logger is enabled only for debug builds and [StrictMode](https://developer.android.com/reference/android/os/StrictMode.html) is enabled for debug as well.


pilhuhn:
lgtm - @danielpassos ?

We may also want to look into automate the build via Travis

arturdryomov:
Yep, Travis and basic `README.md` instructions are coming next.

build.gradle
```diff
@@ -0,0 +1,52 @@
...
+        mavenCentral()
+    }
+
+    dependencies {
+        classpath "com.android.tools.build:gradle:1.2.3"
+        classpath "com.jakewharton.sdkmanager:gradle-plugin:0.12.0"
+    }
+}
+
+apply plugin: "android-sdk-manager"
```
danielpassos, line 12: Any special reason to use it?

build.gradle
```diff
@@ -0,0 +1,52 @@
...
+dependencies {
+    repositories {
+        mavenCentral()
+    }
+
+    compile "com.android.support:support-annotations:22.1.1"
+    compile "com.jakewharton:butterknife:6.1.0"
+    compile "com.jakewharton.timber:timber:3.1.0"
+    compile "com.squareup.dagger:dagger:1.2.2"
+    compile "org.jboss.aerogear:aerogear-android-pipe:2.1.0"
```
danielpassos, line 48: I do not like to use dependency injection on Android, then I would start without it. If in the future we really need it to a good readability of the code, we can talk about some options.

src/main/java/org/hawkular/client/android/backend/BackendClient.java
```diff
@@ -0,0 +1,58 @@
...
+
+	private void setUpPipes() {
+		PipeManager.config(Pipes.TENANTS, RestfulPipeConfiguration.class)
+			.withUrl(getPipeUrl(PipePaths.TENANTS))
+			.forClass(Tenant.class);
+	}
+
+	private URL getPipeUrl(String pipePath) {
+		try {
+			return new URL(new URL(serverUrl), pipePath);
```
danielpassos: We have a utility for it in Pipe

```
UrlUtils.appendToBaseURL(serverUrl, pipePath);
```

src/main/java/org/hawkular/client/android/backend/BackendClient.java
```diff
@@ -0,0 +1,58 @@
...
+	{
+		private PipePaths() {
+		}
+
+		public static final String TENANTS = "hawkular-metrics/tenants";
+	}
+
+	private String serverUrl;
+
+	public void setServerUrl(@NonNull String serverUrl) {
```
danielpassos: I think we can move it for the constructor

src/main/java/org/hawkular/client/android/backend/BackendClient.java
```diff
@@ -0,0 +1,58 @@
...
+import org.jboss.aerogear.android.pipe.LoaderPipe;
+import org.jboss.aerogear.android.pipe.PipeManager;
+import org.jboss.aerogear.android.pipe.rest.RestfulPipeConfiguration;
+
+import java.net.MalformedURLException;
+import java.net.URL;
+
+public final class BackendClient
+{
+	public static final class Pipes
```
danielpassos, line 17: The idea is create a new class for every feature/endpoint?

src/main/res/layout/activity_launcher.xml
```diff
@@ -0,0 +1,24 @@
...
+        android:layout_width="match_parent"
+        android:layout_height="16dp"/>
+
+    <Button
+        android:id="@+id/button_fetch_tenants"
+        android:text="Fetch tenants"
+        android:layout_width="match_parent"
+        android:layout_height="wrap_content"/>
+
+</LinearLayout>
```
danielpassos: Please add a newline at end of file

build.gradle
arturdryomov, line 12: Basically it will download SDK and all required SDK-related components such as required build tools and support library automatically. [Details](https://github.com/JakeWharton/sdk-manager-plugin/blob/master/README.md).

build.gradle
arturdryomov, line 48: Not a problem, but may I ask is there any particular reasoning? If you are worried about performance — almost everything is done [at compile time](http://square.github.io/dagger/).

src/main/java/org/hawkular/client/android/backend/BackendClient.java
arturdryomov: As far as I see it user will be able to change server URL while running the application.

My initial idea was to make the `BackendClient` as a singleton so it can have state across activities. Otherwise it should be recreated by every activity and, as a result, all pipes will be recreated.

src/main/java/org/hawkular/client/android/backend/BackendClient.java
arturdryomov, line 17: Not sure what do you mean, but I thought about storing pipes identifiers and server paths for every  endpoint.

src/main/java/org/hawkular/client/android/backend/BackendClient.java
pilhuhn: The server base URL should be changeable at runtime. Individual sub-urls probably not so.

build.gradle
danielpassos, line 48: The problem is not the performance, but the fact I can't see any benefit in use it instead of don't need to do some findViewById

src/main/java/org/hawkular/client/android/backend/BackendClient.java
danielpassos: My idea was about create a instance of it per server the user want monitor

src/main/java/org/hawkular/client/android/backend/BackendClient.java
danielpassos, line 17: Chillax, after commented I got your point

build.gradle
arturdryomov, line 48: Do you mean Butter Knife or Dagger? Butter Knife is not a dependency injection, it is just an automatic listeners generation.

danielpassos, line 48: Dagger

arturdryomov:
@danielpassos All set.

danielpassos:
:+1:

PR #6, opened 2015-06-03
Resource types screen

arturdryomov:
I tried `RecyclerView` and it remains totally gross regarding simple things — there is no simple ability to add list dividers or just set an item click listener.

Not sure if `BackendClient` should be a singleton really — AeroGear components handle most of the state: `PipeManager`, `AuthorizationManager`. All these `ReadFilter`-related things are potentially not very nice — I had a plan to declare all paths at one place but probably it is not possible. I’ll think about it.

### DIFF
--- a/src/main/java/org/hawkular/client/android/util/Intents.java
+++ b/src/main/java/org/hawkular/client/android/util/Intents.java
@@ -1,0 +1,42 @@
+package org.hawkular.client.android.util;
+
+import android.content.Context;
+import android.content.Intent;
+import android.support.annotation.NonNull;
+
+import org.hawkular.client.android.activity.ResourceTypesActivity;
+import org.hawkular.client.android.backend.model.Tenant;
+
+public class Intents
+{
+	private Intents() {
+	}
+
+	public static final class Extras
+	{
+		private Extras() {
+		}
+
+		public static final String TENANT = "tenant";
+	}
+
+	public static final class Builder
+	{
+		private final Context context;
+
+		public static Builder of(@NonNull Context context) {
+			return new Builder(context);
+		}
+
+		private Builder(Context context) {
+			this.context = context;
+		}
+
+		public Intent buildResourceTypesIntent(@NonNull Tenant tenant) {
+			Intent intent = new Intent(context, ResourceTypesActivity.class);
+			intent.putExtra(Extras.TENANT, tenant);
+
+			return intent;
+		}
+	}
+}
--- a/src/main/AndroidManifest.xml
+++ b/src/main/AndroidManifest.xml
@@ -23,6 +23,10 @@
 
 		</activity>
 
+		<activity
+			android:name=".activity.ResourceTypesActivity">
+		</activity>
+
 		<service
 			android:name="org.jboss.aerogear.android.authorization.oauth2.OAuth2AuthzService">
 		</service>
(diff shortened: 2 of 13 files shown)


danielpassos:
> I tried `RecyclerView` and it remains totally gross regarding simple things — there is no simple ability to add list dividers or just set an item click listener.

Yes, there is. We can create a [RecyclerView.OnItemTouchListener](http://developer.android.com/reference/android/support/v7/widget/RecyclerView.OnItemTouchListener.html) to do the same what [ListView.OnItemClickListner](https://developer.android.com/reference/android/widget/AdapterView.OnItemClickListener.html) do.

See my gist: https://gist.github.com/danielpassos/6aec0da5f53679ae23b6

danielpassos:
> Not sure if `BackendClient` should be a singleton really — AeroGear components handle most of the state: `PipeManager`, `AuthorizationManager`. 

Yeap, you don't need to worry about your Pipe, Auth, etc, created.

danielpassos:
> All these `ReadFilter`-related things are potentially not very nice — I had a plan to declare all paths at one place but probably it is not possible. I’ll think about it.

I'm not following you here. What it means? What do you need in `ReadFilter`?

danielpassos:
I think you can create/store all Pipes, Auths you need in `Application.onCreate` instead of use `BackendClient`, but let's do it step by step.

arturdryomov:
> Yes, there is. We can create a `RecyclerView.OnItemTouchListener` to do the same what `ListView.OnItemClickListener` do.

Thanks, but I know it is possible, it is just not justified to abandon `ListView`, especially when no benefits of `RecyclerView` will be used (probably).

> I'm not following you here. What it means? What do you need in `ReadFilter`?

Well, I thought about putting all paths into the `BackendPipes.Paths` class. But Hawkular paths require variable components, i. e. `/inventory/{tenantId}/resourceTypes`. It is not a great idea to create a new pipe for every one of them, i. e. `/inventory/xxxx-xxxx/resourceTypes` and `/inventory/yyyy-yyyy/resourceTypes`. `ReadFilter` comes in play, but it blocks my idea about all paths in one place. Maybe I should workaround using another way.
- Register pipes with a common path via `PipeManager.config`. For example, tenants, resources and metrics are placed under `https://server.link/hawkular/inventory`, so it will be used with `PipeManager.config` for tenants, resources and metrics as well.
- Fetch pipes using `ReadFilter`. For example, tenants — `tenant/`, resource types — `{tenantId}/resourceTypes`, metrics — `{tenantId}/resourceTypes/{resourceType}`.

What do you think?

> I think you can create/store all Pipes, Auths you need in `Application.onCreate` instead of use `BackendClient,` but let's do it step by step.

Not exactly — the endpoint should be able to change at runtime, so some sort of re-registration should be available.

arturdryomov:
Ping @danielpassos.

src/main/java/org/hawkular/client/android/util/Intents.java
```diff
@@ -0,0 +1,42 @@
...
+			return new Builder(context);
+		}
+
+		private Builder(Context context) {
+			this.context = context;
+		}
+
+		public Intent buildResourceTypesIntent(@NonNull Tenant tenant) {
+			Intent intent = new Intent(context, ResourceTypesActivity.class);
+			intent.putExtra(Extras.TENANT, tenant);
```
danielpassos, line 37: The builder is only to create a intent for tenant?

arturdryomov, line 37: I have a plan to have all intents there and names for extras as well.

PR #8, opened 2015-06-10
License headers fixed or added using mvn license:format

ppalaga:


ppalaga:
Ah, it is failing because Travis relies in maven central only. Let me fix it.

.gitignore
```diff
@@ -7,7 +7,16 @@ local.properties
 .idea/
 *.iml
 
+# Eclipse
```
arturdryomov, line 8: Not sure we should keep it for Studio-based project, but if it is more comfortable for you — not a problem.

.gitignore
```diff
@@ -7,7 +7,16 @@ local.properties
...
+.classpath
+.settings
+
+
 # Gradle
 
 .gradle/
 build/
+
+# Maven
```
arturdryomov, line 17: Nit. An empty line after the header.

arturdryomov:
These directories and files should be ignored, because they are auto-generated.
- `gradle`
- `gradlew`
- `gradlew.bat`
- `*.properties` — not in the repository, but are generated after building a project.

arturdryomov:
@ppalaga Should I fix checkstyle warnings in your branch somehow or just merge and fix afterwards?

arturdryomov:
Another two nits.
- Should the licence header be followed by an empty line like [there](https://github.com/hawkular/hawkular-inventory/blob/7ae4039bea8c82e96ff4516ab75178dd7eb2e27a/api/src/main/java/org/hawkular/inventory/api/InventoryException.java#L17)?
- Should I use `mvn -s .travis.maven.settings.xml verify` locally as well?

.gitignore
ppalaga, line 8: Not sure what Studio-based should imply here... Is this not a usual Gradle project that can be handled by any Gradle-compatible tool incl. Eclipse?

arturdryomov, line 8: As far as I know Eclipse doesn’t support it — at least yet. You can keep, not a big deal.

.gitignore
ppalaga, line 17: You mean the an empty line after the header is missing? I'd say that there is an empty line too much after all other section markers :)

arturdryomov, line 17: That’s why it is nit!

ppalaga:
> These directories and files should be ignored, because they are auto-generated.
> gradle gradlew gradlew.bat *.properties — not in the repository, but are generated after building a project.

Hm... some of them were in git even before my changes. Let me try to remove them.

> Should the licence header be followed by an empty line?

If `mvn license:format` did without an empty line, I'd say that's correct.

> Should I use mvn -s .travis.maven.settings.xml verify locally as well?

Yes, you can, but when thinking of it again, it downloads a lot of maven stuff that is not necessary for the license plugin  and checkstyle. Let me try to find a more sparse form of the mvn command.

arturdryomov:
> Hm... some of them were in git even before my changes. Let me try to remove them.

I meant ignored by checking plugins.

ppalaga:
> > Hm... some of them were in git even before my changes. Let me try to remove them.
> 
> I meant ignored by checking plugins.

OK, I excluded them from license checks https://github.com/ppalaga/android-client/commit/edac4f7b5b197dca948d13b9160211391e36b36e I hope the list is complete

ppalaga:
The last commit is a result of bulk Source > Format operation in Eclipse. Not sure if this will suit Android Studio. Feel free to criticize or propose something else. Anyway, Checkstyle is passing like that. https://github.com/hawkular/android-client/commit/13c52e0538d73b6cf39521756a43ebd4587fdcba

ppalaga:
> Should I use mvn -s .travis.maven.settings.xml verify locally as well?

I changed to `mvn license:check checkstyle:check` However, I am not 100% sure it saved any bandwith.

arturdryomov:
marvelous

@ppalaga Is the `.travis.maven.settings.xml` really necessary though? Seems like it just adds repositories and the `hawkular-parent` is available at Central. Isn’t it better to move these declarations to the parent POM directly?

ppalaga:
@ming13 yes, `.travis.maven.settings.xml` makes sense, because `hawkular-parent` reaches maven central within 24hrs from the release to JBoss Nexus and I typically upgrade parent in consuming projects just after releasing to nexus.

arturdryomov:
@ppalaga, thanks for your work!

PR #12, opened 2015-06-13
Metric chart

arturdryomov:
device-2015-06-14-024505

### DIFF
--- a/src/main/java/org/hawkular/client/android/activity/MetricDataActivity.java
+++ b/src/main/java/org/hawkular/client/android/activity/MetricDataActivity.java
@@ -33,6 +33,7 @@ import org.jboss.aerogear.android.pipe.callback.AbstractActivityCallback;
 import java.text.DateFormat;
 import java.util.ArrayList;
 import java.util.Collections;
+import java.util.Comparator;
 import java.util.Date;
 import java.util.List;
 
@@ -95,6 +96,8 @@ public class MetricDataActivity extends AppCompatActivity {
     }
 
     private void setUpMetricData(List<MetricData> metricDataList) {
+        sortMetricData(metricDataList);
+
         List<PointValue> chartPoints = new ArrayList<>();
 
         List<AxisValue> chartAxisPoints = new ArrayList<>();
@@ -133,6 +136,10 @@ public class MetricDataActivity extends AppCompatActivity {
         hideProgress();
     }
 
+    private void sortMetricData(List<MetricData> metricDataList) {
+        Collections.sort(metricDataList, new MetricDataComparator());
+    }
+
     private void hideProgress() {
         ViewDirector.of(this).using(R.id.animator).show(R.id.chart);
     }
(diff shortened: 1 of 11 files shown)


arturdryomov:
CC @pilhuhn @danielpassos

src/main/java/org/hawkular/client/android/activity/MetricDataActivity.java
```diff
@@ -0,0 +1,167 @@
...
+
+        Viewport chartViewport = new Viewport(chart.getMaximumViewport());
+
+        chartViewport.bottom = chart.getMaximumViewport().bottom - 50;
+        chartViewport.top = chart.getMaximumViewport().top + 50;
+
+        chart.setMaximumViewport(chartViewport);
+        chart.setCurrentViewport(chartViewport);
+
+        hideProgress();
```
pilhuhn, line 136: Just curious - where is the corresponding `startProgress()` ?

src/main/java/org/hawkular/client/android/activity/MetricDataActivity.java
```diff
@@ -0,0 +1,167 @@
...
+
+            chartPoints.add(new PointValue(metricDataPosition, metricData.getValue()));
+
+            chartAxisPoints.add(new AxisValue(metricDataPosition)
+                .setLabel(DateFormat.getTimeInstance(DateFormat.SHORT).format(new Date(metricData.getTimestamp()))));
+        }
+
+        Line chartLine = new Line(chartPoints)
+            .setColor(getResources().getColor(R.color.background_primary_dark))
+            .setCubic(true)
```
pilhuhn, line 116: This is probably how the values are interpolated and so on?
Do you think it could make sense later to expose that to the user to play with?

src/main/java/org/hawkular/client/android/activity/MetricDataActivity.java
```diff
@@ -0,0 +1,167 @@
...
+    private Metric getMetric() {
+        return getIntent().getParcelableExtra(Intents.Extras.METRIC);
+    }
+
+    private void setUpMetricData(List<MetricData> metricDataList) {
+        List<PointValue> chartPoints = new ArrayList<>();
+
+        List<AxisValue> chartAxisPoints = new ArrayList<>();
+
+        for (int metricDataPosition = 0; metricDataPosition < metricDataList.size(); metricDataPosition++) {
```
pilhuhn, line 105: In the screen shot, the order (by time) seems to be reversed (02:44 at the left, 02:35 on the right).
You should perhaps sort the data here, as you should probably not be sure that the server always returns data sorted by timestamp.

src/main/java/org/hawkular/client/android/activity/MetricDataActivity.java
```diff
@@ -0,0 +1,167 @@
...
+        chartViewport.top = chart.getMaximumViewport().top + 50;
+
+        chart.setMaximumViewport(chartViewport);
+        chart.setCurrentViewport(chartViewport);
+
+        hideProgress();
+    }
+
+    private void hideProgress() {
+        ViewDirector.of(this).using(R.id.animator).show(R.id.chart);
```
pilhuhn, line 144: Oh, I thought hideProgress hides some spinner -- with that impl, perhaps change the name to `showChart()`?

src/main/java/org/hawkular/client/android/activity/MetricDataActivity.java
arturdryomov, line 136: It is named as `MetricDataActivity#showProgress`.

src/main/java/org/hawkular/client/android/activity/MetricDataActivity.java
arturdryomov, line 116: Hmm, honestly don’t know. I tried to mimic the current web UI behaviour as much as possible. Maybe dots with labels can be useful, but it looks kind of messy, especially in the portrait orientation. Scrolling backwards will be really useful, but I didn’t get it work with the current library at this point.

arturdryomov:
@pilhuhn Sorted!

device-2015-06-14-130946

The chart itself will need a smarter approach in the future. For some reason putting chart points as `(timestamp, value)` breaks the chart, so I was forced to use a simpler technique. As a result you can see that the bottom axis has duplicate values.

PR #15, opened 2015-06-16
Alert push notifications

arturdryomov:
After a very tedious searching for solutions it is mostly done. The Hawkular part is still work in the progress — message payload is not very useful. It will be changed in the future.

```
Alert [alertId=a3ca17295646a583974e9f1c01737028_trigger_thres|1434447387356, status=OPEN, ackTime=0, ackBy=null, resolvedTime=0, resolvedBy=null]
```

I combined [a guide](https://github.com/hawkular/hawkular-android-client/wiki/Push-Notifications) describing what to do step-by-step to activate notifications. Actually you don’t need to configure Hawkular AeroGear UPS plugin at all — it is possible to send notifications via UPS console. But it works if you want to.

I tried to make push notifications optional — if you don’t configure notifications the application will work simply without notifications.

device-2015-06-16-142939


arturdryomov:
CC @pilhuhn @danielpassos

danielpassos:
Could register device and receive a message

PR #19, opened 2015-06-20
Refactoring

arturdryomov:
Things! Excitement!

Seems big, but surprisingly minor changes. Mostly fixes for what I always wanted to fix. I tried to make commits more or less atomic. [Commit messages are not generated](http://whatthecommit.com/), I swear.


src/main/java/org/hawkular/client/android/activity/MetricDataActivity.java
```diff
@@ -80,7 +82,19 @@ private void setUpToolbar() {
...
         showProgress();
 
-        BackendClient.getInstance().getMetricData(getTenant(), getMetric(), this, new MetricDataCallback());
+        BackendClient.of(this).getMetricData(
+            getTenant(), getMetric(), getMetricStartTime(), getMetricFinishTime(), new MetricDataCallback());
+    }
+
+    private Date getMetricStartTime() {
+        Calendar calendar = GregorianCalendar.getInstance();
+        calendar.add(Calendar.MINUTE, -10);
```
pilhuhn, line 91: In the UI the standard interval is currently 1h ( in RHQ we used to have 8h as default) - should the Android client do the same?

src/main/java/org/hawkular/client/android/util/Formatter.java
```diff
@@ -0,0 +1,31 @@
...
+import android.support.annotation.IntRange;
+import android.support.annotation.NonNull;
+import android.text.format.DateUtils;
+
+public final class Formatter {
+    private Formatter() {
+    }
+
+    public static String formatTime(@NonNull Context context, @IntRange(from = 0) long millis) {
+        return DateUtils.formatDateTime(context, millis, DateUtils.FORMAT_SHOW_DATE | DateUtils.FORMAT_SHOW_TIME);
```
pilhuhn, line 29: Does that automagically the locale into account?

src/main/java/org/hawkular/client/android/activity/MetricDataActivity.java
arturdryomov, line 91: Was it changed? I remember it was 10 minutes at the UI, maybe I am wrong. I’ll tweak the chart further as a standalone request anyway, the current chart is available to display only 10 minutes without messing it up.

src/main/java/org/hawkular/client/android/util/Formatter.java
arturdryomov, line 29: Yep, it does.

src/main/java/org/hawkular/client/android/util/Preferences.java
```diff
@@ -40,7 +41,14 @@ private Keys() {
         public static final String BACKEND_ENVIRONMENT = "environment";
     }
 
-    private SharedPreferences serverPreferences;
+    public static final class Defaults {
+        private Defaults() {
+        }
+
+        public static final int BACKEND_PORT = Integer.MIN_VALUE;
```
pilhuhn, line 48: That looks odd .. is no port is given, will the user see a very negative number?
Why not use 18080 or such?

arturdryomov, line 48: Take a look at the `AuthorizationActivity#isPortCorrectNumber`. It is the invalid number used for checking purposes.

PR #24, opened 2015-06-23
Add some DE translation + some explanation on Auth screen.

pilhuhn:


arturdryomov:
Formatting is a bit off, but I’ll fix it later.

PR #26, opened 2015-06-30
Deauthorization

arturdryomov:
GIF to the rescue!

animation


danielpassos:
I like the idea but I think it is not so intuitive for the users. Maybe just move it to the end of the navigation drawer is better

arturdryomov:
@danielpassos, that’s how it is done via almost all Google apps: Play, Gmail, Drive, Photos, you name it. And it is at [the guidelines](http://www.google.com/design/spec/patterns/navigation-drawer.html#navigation-drawer-specs).

danielpassos:
I don't agree with everything Google do, still sucks for me. Anyway in Google apps you don't have an option to logout in navigation drawer, only to go to your account in account manager.

I'm +1 to have an option to switch between servers using that approach (like Google to between accounts) but not to have a logout on it.

secondsun:
I concur with @danielpassos but for slightly differing reasons.  Generally the navigation drawer is a bad place to put destructive actions such as logging out.

arturdryomov:
# Screenshots

## Drive

drive

## Dropbox

dropbox

## Photos

photos

## Pocket Casts

pocket casts

## Reddit Sync

reddit sync

arturdryomov:
But these animations you guys!

I see nothing wrong with the current approach actually. “Manage Accounts” at Google applications just throws user to a system settings mambo-jambo where you need to select an account, then press on it again, then press the overflow button, then press the removal button and then, hopefully, you will get the account removed. Our approach compared to this is just a magical simplicity itself. And... the Hawkular web interface hides logging out at the dropdown of the account name label.

As screenshots show above, first-party applications prefer this approach. Photos even shows removal button right there. Third-party applications prefer to stick to settings. I would say it is “as it is” vs “settings” at this point.

pilhuhn:
@secondsun @danielpassos How can @ming13 move forward here - what do you propose as alternative(s)?

secondsun:
Just show a confirm box when you log out, or better yet navigate to a confirm screen.

I don't mind a navigation to log out from the navigation drawer.  It is just the destructive action of logging out happening from the nav drawer.  It is of course just my opinion and feel free to ignore it.

pilhuhn:
So you say, that the current positioning of this in the drawer is ok, as 
long as an additional
"do you really want to log out? yes/no/perhaps/42" dialog pops up?

arturdryomov:
Related to #34.

Nope, @secondsun, I am not gonna ignore you, don’t count on that ;-)

I propose listing accounts or personas (take a look at #34 again) this way and have an item named like “Sign Out” at the “Account” section of the Settings screen we don’t have yet. This will work better with personas actually.

The question is on the motion, all those in favor say aye. All those opposed say nay and propose alternative as well.

secondsun:
@pilhuhn yup
@ming13 Aye :)

pilhuhn:
I think switching personas is more common than switching between completely independent Hawkular instances.
Having a dedicated account section + logout action sounds good to me.

arturdryomov:
Was superseded by #36. Thanks everyone!

PR #28, opened 2015-07-08
API update

arturdryomov:
Mostly a long awaited update for the Hawkular 1.0.0 α 2. As a bonus, the application supports authorization without port set, mostly to handle `demo.hawkular.org` properly. Yep, it is the default one at this point.

### DIFF
--- a/src/main/java/org/hawkular/client/android/util/Ports.java
+++ b/src/main/java/org/hawkular/client/android/util/Ports.java
@@ -1,0 +1,15 @@
+package org.hawkular.client.android.util;
+
+import android.support.annotation.IntRange;
+
+public final class Ports {
+    private Ports() {
+    }
+
+    public static final int MINIMUM = 0;
+    public static final int MAXIMUM = 0xFFFF;
+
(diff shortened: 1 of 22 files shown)


arturdryomov:
CC @pilhuhn @danielpassos

src/main/java/org/hawkular/client/android/util/Ports.java
```diff
@@ -0,0 +1,31 @@
...
+import android.support.annotation.IntRange;
+
+public final class Ports {
+    private Ports() {
+    }
+
+    public static final int MINIMUM = 0;
+    public static final int MAXIMUM = 0xFFFF;
+
+    public static boolean isCorrect(@IntRange(from = Ports.MINIMUM, to = Ports.MAXIMUM) int port) {
```
pilhuhn, line 28: I generally wonder if we should support non-numeric ports like "https" or "http" instead 443 and 80 - as they show up in /etc/services. But then, the current xml definition only allows numbers, so we should be good.

PR #29, opened 2015-07-12
Alerts time filtering

arturdryomov:
Most likely should be rebased after #28.

device-2015-07-12-232548
device-2015-07-12-232536


arturdryomov:
Rebased and seems to work with 1.0.0 α 2.

Issue #31, opened 2015-07-14
Alerts categorization


arturdryomov:
At this point alerts are shown without any filtering or categorization, i. e. there are no ways to take a look at alerts only for `https://github.com` or `https://reddit.com`, for example,—only everything you got.

A search for solution leads to both API and UI confusion.
### API

It is possible to filter alerts using trigger IDs and tags. Console uses [filtering based on triggers](https://github.com/hawkular/hawkular/blob/640b86adf45ff8f1f38c48bf6f072fb236021a6c/console/src/main/scripts/plugins/metrics/html/url-list.html#L54) which I personally consider as a hack, because there is no guarantee that the trigger ID will have a format of `{resource-id}_trigger` on an infinite timescale.

Anyway, `GET /alerts` result looks [like that](http://pastie.org/private/y47oftsdoi4bixbndpfcw). As you can see, there is not much to use to determine a category, i. e. a resource in our case.

We can work around for time being the same ways as Console does—it is the solution at this point.
### UI

The current scheme [looks like that](https://github.com/hawkular/hawkular-android-client/wiki/Design).
- Navigation Drawer. Contains “Resources” and “Alerts” sections.
- Resources section. Resources View. Contains resources with resource types as sections. Selecting a resource opens Metrics View. Selecting a metric shows a Metric View.
- Alerts section. Alerts view. Contains alerts. Selecting an alert does nothing.

To filter alerts we need to select a resource first.
- Navigation Drawer. Contains “Resources” and “Alerts” sections.
- Resources section. Resources View. Contains resources with resource types as sections. Selecting a resource opens Metrics View. Selecting a metric shows a Metric View.
- Alerts section. Resources View. Contains resources with resource types as sections. Selecting a resource opens Alerts View.

As you can notice the screen with the Navigation Drawer in the second scenario contains the Resources View all the time, which can lead to confusion.

The closes solution I can thing of is showing “Alerts” and “Resources” as a title of the screen with the Navigation Drawer. This probably will give at least a visual hint of current “mode”.

pilhuhn:
I agree that the current api filter by trigger is limiting. I also believe that some kind of search and filtering is being added as we speak. One of the issues is that Alerts is trying to be generic and could be applicable out of the main Hawkular context.

It may be though that you need to query some properties from Inventory before being able to filter on them in alerts.

With respect to the API, I would like you to clearly state what you need, so that this can be added to Hawkular - especially as you will not be the only client, so others will benefit here as well.

[...]

PR #38, opened 2015-08-03
Alerts categorization

arturdryomov:
Reasoning is described at #31.


src/main/java/org/hawkular/client/android/fragment/AlertsFragment.java
```diff
@@ -62,6 +65,10 @@
 
     @State
     @Nullable
+    ArrayList<Trigger> triggers;
```
pilhuhn, line 68: Usually it is good practice to use the interface on the declaration, unless you need that concrete type later (e.g. because it has a method that is not in the interface)

arturdryomov, line 68: I know and follow that rule every time, but Icepick — the library which handles state on configuration changes — doesn’t accept generics, because in that case it doesn’t know how to put the field to a `Parcel` container — which doesn’t support generics as well.

PR #39, opened 2015-08-04
Real alert actions

arturdryomov:
The previous implementation was just an UI element. With Hawkular α 3 it is possible to execute real actions. This request also changes default Alerts loading to fetch only open alerts, ignoring resolved and acknowledged ones.

### DIFF
--- a/src/main/java/org/hawkular/client/android/util/Uris.java
+++ b/src/main/java/org/hawkular/client/android/util/Uris.java
@@ -20,12 +20,21 @@ import android.net.Uri;
 import android.support.annotation.NonNull;
 import android.text.TextUtils;
 
+import java.io.UnsupportedEncodingException;
 import java.net.URI;
 import java.net.URISyntaxException;
+import java.net.URLEncoder;
 import java.util.List;
 import java.util.Map;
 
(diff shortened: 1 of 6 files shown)


src/main/java/org/hawkular/client/android/util/Uris.java
```diff
@@ -20,12 +20,21 @@
...
+import java.net.URLEncoder;
 import java.util.List;
 import java.util.Map;
 
 public final class Uris {
+    private static final class Charsets {
+        private Charsets() {
+        }
+
+        public static final String UTF_8 = "UTF-8";
```
pilhuhn, line 35: Wow, I would have expected that there is a constant already somewhere, but apparently not.

arturdryomov, line 35: Well, it is possible to refer it.

``` java
java.nio.charset.StandardCharsets.UTF_8.name()
```

One catch though—available only from API 19, i. e. KitKat.

pilhuhn, line 35: There is no need to change your code. I was just surprised

PR #40, opened 2015-08-06
Fix imports order to match the new Checstyle rule

ppalaga:


ppalaga:
@ming13 could you please review? Esp. take care of the second commit 392391d  has not broken something - it is a result of fix Organize imports action in my Eclipse, where I had the class path incomplete. So maybe Eclipse has removed imports that it could not resolve. Eventually ignore that commit and reorder the imports using your IDE.

ppalaga:
@ming13 indeed my second commit 392391d did what I said: https://travis-ci.org/hawkular/hawkular-android-client/builds/74364640#L1811 I have removed it from this PR. So please take the upgrade to parent 19 d8eb7fd and reorder the imports using your IDE.

ppalaga:
replaced by #41

PR #43, opened 2015-08-25
Release configuration

arturdryomov:
[Important conversation about DEX methods count](http://lists.jboss.org/pipermail/aerogear-dev/2015-August/012000.html).
- ProGuard to strip unused code.
- Resource shrinking.
- Signing.

ProGuard applied to debug builds as well to avoid [multidex](https://developer.android.com/tools/building/multidex.html).

parks-checkmate

Fun facts.
- Methods count.
  - Debug: before: 65346
  - Debug: after: 17732
  - Release: 17731
- APK size.
  - Debug: before: 4.1 MB
  - Debug: after: 1.6 MB
  - Release: 1.3 MB
- Build time.
  - Debug: before: 100 sec.
  - Debug: after: 40 sec.
  - Release: 45 sec.

The speed difference is there because the conversion from bytecode to DEX takes the longest time during the build. Less code—less conversion to DEX.

And yes, I need a MacBook Pro upgrade to speed up the build, but I am awating for Skylake like a fool.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,27 +1,37 @@
-# Android SDK
+# Configuration
+
+## Android SDK
 
 local.properties
 
-# Android Studio
+## Server
+
+/server.properties
+
+## Signing
+
+/signing.properties
+
+# IDE
+
+## Android Studio
 
 .idea/
 *.iml
 
-# Eclipse
+## Eclipse
 
 .project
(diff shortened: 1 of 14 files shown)


danielpassos:
Approval intense man clapping

PR #48, opened 2016-02-10
Added login with QR and Auth system changed from access token to key/secret

openwarrior:
Intent integration is done for calling Scanner application of ZXing.
Auth system is changed from access token to key secret pair, to remove dependency from keycloak.
Implemented AuthzModule is used instead of complete Aerogear authz to sustain connectivity with pipe. 
Changes still needed to be commented in better way. Working on that

### DIFF
--- a/src/main/AndroidManifest.xml
+++ b/src/main/AndroidManifest.xml
@@ -1,7 +1,6 @@
-<?xml version="1.0" encoding="utf-8"?>
 <!--
 
-    Copyright 2015 Red Hat, Inc. and/or its affiliates
+    Copyright 2015-2016 Red Hat, Inc. and/or its affiliates
     and other contributors as indicated by the @author tags.
 
     Licensed under the Apache License, Version 2.0 (the "License");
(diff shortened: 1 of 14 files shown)


src/main/res/values/strings.xml
```diff
@@ -1,7 +1,6 @@
-<?xml version="1.0" encoding="utf-8"?>
```
arturdryomov, line 1: Don’t think it was necessary to reformat everything in this file for this change.

src/main/AndroidManifest.xml
```diff
@@ -1,7 +1,6 @@
-<?xml version="1.0" encoding="utf-8"?>
```
arturdryomov, line 1: Don’t think it was necessary to reformat everything in this file for this change.

openwarrior, line 1: actually it was changed by the formatting command

src/main/java/com/google/zxing/integration/android/IntentIntegrator.java
```diff
@@ -0,0 +1,506 @@
+/*
```
arturdryomov, line 1: Please use Gradle dependency instead of pulling in unnecessary third-party code.

```
compile 'com.google.zxing:android-integration:3.2.1'
```

Also, have you considered [GPS barcode scanner](http://android-developers.blogspot.com.by/2015/08/barcode-detection-in-google-play.html)? Most likely it will remove dependency on third-party application for scanning.

src/main/java/com/google/zxing/integration/android/IntentResult.java
```diff
@@ -0,0 +1,90 @@
+/*
```
arturdryomov, line 1: Please use Gradle dependency instead of pulling in unnecessary third-party code.

```
compile 'com.google.zxing:android-integration:3.2.1'
```

Also, have you considered [GPS barcode scanner](http://android-developers.blogspot.com.by/2015/08/barcode-detection-in-google-play.html)? Most likely it will remove dependency on third-party application for scanning.

openwarrior:
What is the problem this time.
I formatted with the mvn license:format

arturdryomov:
You have to learn to read Travis log, it speaks for itself most of the time. `license:format` only fixes licence headers. You’ve reformatted XML files for some reason and checkstyle doesn’t like that.

src/main/java/com/google/zxing/integration/android/IntentResult.java
openwarrior, line 1: Ok. using gradle dependency now. but using the ZXing app was suggested will by Juca. Will try with  GPS barcode scanner too

pilhuhn:
Thank you @ming13 for checking.
@anuj1708 There are templates for Eclipse and IntelliJ that should set default format settings that are supposed to be compatible with checkstyle. Have a look here: https://github.com/hawkular/hawkular-build-tools/tree/master/ide-configs

PR #50, opened 2016-02-15
Provide Auth system for using key secret pair

openwarrior:
At present this client use  Access Tokens which restrict use of secret store.
To be able to use secret store by providing minimum changes AuthzModule
is Implemented. It performs all related work of generating new key and secret pair, Storing them as account, retrieving stored account, deleting stored account. 

ModifiedAuthzModule - Implement AuthzModule from Aerogear Authz
By this rest of connectivity remain same.

ModuleKeeper - Retain Modified AuthzModule in a Map in the imitation of
AuthorizationManager


src/main/java/org/hawkular/client/android/auth/ModifiedAuthzModule.java
```diff
@@ -0,0 +1,197 @@
...
+import android.util.Log;
+
+/**
+ * ModifiedAuthzModule.
+ * <p/>
+ * Performs all related work of generating new key and secret pair, Storing them as account,
+ * Retrieving stored account, Deleting stored account.
+ */
+
+public class ModifiedAuthzModule implements AuthzModule {
```
pilhuhn: There should be a better class name like `HawkularAuthzModule` or `OAuth2AuthzModule` or such -- depending on what the addition to the base module is. Or perhaps `SecretStoreAuthzModule` -- @jpkrohling perhaps also has an idea.

jpkrohling: +1 for `SecretStoreAuthzModule`

src/main/java/org/hawkular/client/android/auth/ModifiedAuthzModule.java
```diff
@@ -0,0 +1,197 @@
...
+
+    Context applicationContext;
```
arturdryomov: Can (and should be) `private` and `final`.

src/main/java/org/hawkular/client/android/auth/ModifiedAuthzModule.java
```diff
@@ -0,0 +1,197 @@
...
+    private SQLStore<OAuth2AuthzSession> sessionStore;
```
arturdryomov: Can be `final`.

src/main/java/org/hawkular/client/android/auth/ModifiedAuthzModule.java
```diff
@@ -0,0 +1,197 @@
...
+
+    public ModifiedAuthzModule(Context context) {
+        applicationContext = context;
```
arturdryomov: `applicationContext` can be renamed to `context`. Also, prefer this — `this.context = context.getApplicationContext()`.

src/main/java/org/hawkular/client/android/auth/ModifiedAuthzModule.java
```diff
@@ -0,0 +1,197 @@
...
+    }
+
+    @Override
+    public boolean isAuthorized() {
+        return true;
+    }
+
+    @Override
+    public boolean hasCredentials() {
+        OAuth2AuthzSession storedAccount = sessionStore.read("hawkular");
```
arturdryomov: Please don’t ever, ever, ever use hardcoded text. Move it to a constant.

src/main/java/org/hawkular/client/android/auth/ModifiedAuthzModule.java
```diff
@@ -0,0 +1,197 @@
...
+        if (storedAccount != null) {
+            return true;
+        }
+        return false;
```
arturdryomov: Can be simplified to `storedAccount != null`.

src/main/java/org/hawkular/client/android/auth/ModifiedAuthzModule.java
```diff
@@ -0,0 +1,197 @@
...
+
+        sessionStore = (SQLStore<OAuth2AuthzSession>) DataManager.getStore("sessionStore");
+        sessionStore.openSync();
+        return sessionStore;
+    }
+
+    public void addAccount(OAuth2AuthzSession session) {
+        sessionStore.save(session);
+    }
+
```
arturdryomov: Redundant empty lines.

src/main/java/org/hawkular/client/android/auth/ModifiedAuthzModule.java
openwarrior: SecretStoreAuthzModule got nominated  :smile:

src/main/java/org/hawkular/client/android/auth/ModifiedAuthzModule.java
openwarrior: That right thanks

src/main/java/org/hawkular/client/android/auth/ModifiedAuthzModule.java
openwarrior: done

src/main/java/org/hawkular/client/android/auth/ModifiedAuthzModule.java
openwarrior: turning name to context. But as i am already sending applicationContext to it. Can't find need of 
this.context = context.getApplicationContext().

src/main/java/org/hawkular/client/android/auth/ModifiedAuthzModule.java
openwarrior: Note is taken :)

src/main/java/org/hawkular/client/android/auth/ModifiedAuthzModule.java
openwarrior: Was having a sense that doing something Dumb

src/main/java/org/hawkular/client/android/auth/ModifiedAuthzModule.java
openwarrior: Always get confused with this. Can you please point me to a convention about it.

src/main/java/org/hawkular/client/android/auth/ModifiedAuthzModule.java
```diff
@@ -0,0 +1,197 @@
...
+        moduleFields.setQueryParameters(authzFields.getQueryParameters());
+        return moduleFields;
+    }
+
+    @Override
+    public boolean handleError(HttpException exception) {
+        return false;
+    }
+
+    private SQLStore<OAuth2AuthzSession> openSessionStore() {
```
arturdryomov: Try to place methods in usage order, i. e. this one should be after the constructor.

src/main/java/org/hawkular/client/android/auth/ModifiedAuthzModule.java
```diff
@@ -0,0 +1,197 @@
...
+            OAuth2AuthzSession storedAccount = sessionStore.read("hawkular");
+            final String token = storedAccount.getAccessToken();
+            try {
+                callback.onSuccess(token);
+            } catch (Exception e) {
+                callback.onFailure(e);
+            }
+        } else {
+
+            AsyncTask<Void, Void, HeaderAndBody> task = new AsyncTask<Void, Void, HeaderAndBody>() {
```
arturdryomov: Most likely you can get away with a simple thread or single threaded executor. This way you can avoid storing exceptions in fields.

src/main/java/org/hawkular/client/android/auth/ModifiedAuthzModule.java
```diff
@@ -0,0 +1,197 @@
...
+                private Exception exception;
+                String json;
+
+                @Override
+                protected HeaderAndBody doInBackground(Void... params) {
+                    HeaderAndBody result;
+                    try {
+                        HttpProvider provider =
+                                new HttpRestProvider(new URL(((LoginActivity) (activity)).backendUrl.toString()
```
arturdryomov: This casting is bad. Your class should be as abstract as possible allowing any activity to use it.

src/main/java/org/hawkular/client/android/auth/ModifiedAuthzModule.java
```diff
@@ -0,0 +1,197 @@
...
+                                        + "/secret-store/v1/tokens/create"));
+                        provider.setDefaultHeader("Accept", "application/json");
+                        provider.setDefaultHeader("Authorization", "Basic "
+                                + buildLoginData(((LoginActivity) (activity)).username,
+                                ((LoginActivity) (activity)).password));
+                        result = provider.post("");
+                        Log.d("Called with POST to: ", new String(provider.getUrl().toString()));
```
arturdryomov: Use `Timber`. It will disable logging for release builds for you.

src/main/java/org/hawkular/client/android/auth/ModifiedAuthzModule.java
arturdryomov: You shouldn’t rely on any conventions when building API. Enforce all you can instead of relying on rules. In this case it is simple, safe and cheap.

src/main/java/org/hawkular/client/android/auth/ModifiedAuthzModule.java
arturdryomov: Just have a single line between methods. As far as I know Hawkular doesn’t have any strict rules except Checstyle ones. You can take a look at [Google ones](https://google.github.io/styleguide/javaguide.html) for some motivation and inspiration.

openwarrior:
@ming13 @pilhuhn @jpkrohling  @danielpassos  Can you please review again. Applied approx everything except Thread coz wanted to leave more like cookbook code of aerogear. And I don't know if line redundancy is upto the standard or not but i tried.

src/main/java/org/hawkular/client/android/auth/SecretStoreAuthzModule.java
```diff
@@ -0,0 +1,185 @@
...
+    }
+
+    @Override
+    public boolean isAuthorized() {
+        return true;
+    }
+
+    @Override
+    public boolean hasCredentials() {
+        OAuth2AuthzSession storedAccount = sessionStore.read(AuthEndPoints.NAME);
```
arturdryomov: Let’s simplify once again ;-) `return sessionStore.read(AuthEndpoints.NAME) != null`

src/main/java/org/hawkular/client/android/auth/SecretStoreAuthzModule.java
```diff
@@ -0,0 +1,185 @@
...
+/**
+ * ModifiedAuthzModule.
+ * <p/>
+ * Performs all related work of generating new key and secret pair, Storing them as account,
+ * Retrieving stored account, Deleting stored account.
+ */
+
+public class SecretStoreAuthzModule implements AuthzModule {
+
+    private Context context;
```
arturdryomov: Still not `final` for both fields. The empty line before them is redundant as well (sorry).

src/main/java/org/hawkular/client/android/auth/SecretStoreAuthzModule.java
```diff
@@ -0,0 +1,185 @@
...
+                HeaderAndBody result;
+                try {
+                    HttpProvider provider =
+                            new HttpRestProvider(new URL(url.toString()
+                                    + AuthEndPoints.Endpoints.ACCESS));
+                    provider.setDefaultHeader("Accept", "application/json");
+                    provider.setDefaultHeader("Authorization", "Basic "
+                            + buildLoginData(username, password));
+                    result = provider.post("");
+                    Timber.e("Called with POST to: ", new String(provider.getUrl().toString()));
```
jpkrohling: `Timber.e` -> sounds like "e" stands for "error", no? Shouldn't it be `Timber.d`?

src/main/java/org/hawkular/client/android/auth/SecretStoreAuthzModule.java
```diff
@@ -0,0 +1,185 @@
...
+ */
+
+public class SecretStoreAuthzModule implements AuthzModule {
+
+    private Context context;
+    private SQLStore<OAuth2AuthzSession> sessionStore;
+
+    public SecretStoreAuthzModule(Context context) {
+        this.context = context.getApplicationContext();
+        sessionStore = openSessionStore();
```
arturdryomov: Personally I prefer to set all fields using `this.` in constructors, but you call the shots.

src/main/java/org/hawkular/client/android/auth/SecretStoreAuthzModule.java
```diff
@@ -0,0 +1,185 @@
...
+    @Override
+    public boolean hasCredentials() {
+        OAuth2AuthzSession storedAccount = sessionStore.read(AuthEndPoints.NAME);
+        return storedAccount != null;
+    }
+
+    public void requestAccess(final Activity activity, final Callback<String> callback,
+                              final String username, final String password, final URL url) {
+
+        Exception exception;
```
arturdryomov: I’m pretty sure that these two variables are useless, because they are duplicated in the inner class.

src/main/java/org/hawkular/client/android/auth/SecretStoreAuthzModule.java
```diff
@@ -0,0 +1,185 @@
...
+                }
+            }
+        }.executeOnExecutor(AsyncTask.THREAD_POOL_EXECUTOR);
+    }
+
+
+    @Override
+    public void requestAccess(final Activity activity, final Callback<String> callback) {
+        OAuth2AuthzSession storedAccount = sessionStore.read(AuthEndPoints.NAME);
+        final String token = storedAccount.getAccessToken();
```
arturdryomov: `final` is useless here. Also, consider this.

``` java
try {
    callback.onSuccess(sessionStore.read(AuthEndpoints.NAME).getAccessToken());
} catch (Exception e) {
    callback.onFailure(e);
}
```

src/main/java/org/hawkular/client/android/auth/SecretStoreAuthzModule.java
```diff
@@ -0,0 +1,185 @@
...
+        try {
+            callback.onSuccess(token);
+        } catch (Exception e) {
```
arturdryomov: Generally it is a bad practice to catch every exception there is. Is there any reasoning behind this?

src/main/java/org/hawkular/client/android/auth/SecretStoreAuthzModule.java
arturdryomov: Yep, `Timber.e` indicates error level posted to logcat.

src/main/java/org/hawkular/client/android/auth/SecretStoreAuthzModule.java
```diff
@@ -0,0 +1,185 @@
...
+                    HttpProvider provider =
+                            new HttpRestProvider(new URL(url.toString()
+                                    + AuthEndPoints.Endpoints.ACCESS));
+                    provider.setDefaultHeader("Accept", "application/json");
+                    provider.setDefaultHeader("Authorization", "Basic "
+                            + buildLoginData(username, password));
+                    result = provider.post("");
+                    Timber.e("Called with POST to: ", new String(provider.getUrl().toString()));
+                    Timber.e("Result of the POST: ", new String(result.getBody()));
+                    json = new String(result.getBody());
```
arturdryomov: As far as I see it you don’t need `json` variable at all.

src/main/java/org/hawkular/client/android/auth/AuthEndPoints.java
```diff
@@ -0,0 +1,30 @@
...
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.hawkular.client.android.auth;
+
+public final class AuthEndPoints {
```
arturdryomov: Pretty sure it is a single word `Endpoints`.

src/main/java/org/hawkular/client/android/auth/AuthEndPoints.java
```diff
@@ -0,0 +1,30 @@
...
+    public static final String NAME = "hawkular";
+    public static final String STORE = "sessionStore";
+
+    public static final class Endpoints {
```
arturdryomov: At this point it is mostly paths, not endpoints.

src/main/java/org/hawkular/client/android/auth/SecretStoreAuthzModule.java
```diff
@@ -0,0 +1,185 @@
...
+        return true;
+    }
+
+    @Override
+    public boolean hasCredentials() {
+        OAuth2AuthzSession storedAccount = sessionStore.read(AuthEndPoints.NAME);
+        return storedAccount != null;
+    }
+
+    public void requestAccess(final Activity activity, final Callback<String> callback,
```
arturdryomov: Suddenly it seems like you don’t need `activity` after all.

src/main/java/org/hawkular/client/android/auth/SecretStoreAuthzModule.java
openwarrior: Can't know list of multiple exception which may arise. May be caused less of experience.
How to know. If net disconnect, wrong URL (that doesn't have hawkular), wrong password. And many more.

jpkrohling: It's a good practice to catch only exceptions you care about, ie, those that you are able to handle. Catching the generic "Exception" means you are handling everything. In general, only code that interfaces with the end user needs to have such generic catches.

On this case, you could handle some scenarios: if the connection fails mid-flight, you might want to retry it. If you got a 404, you might want to give up immediately, as the outcome won't change anyway. Similarly, if you got a "Forbidden", it might mean that the user's password is wrong. For all cases you didn't predict, let it explode until some upper layer catches it and handle it appropriately (for instance, asking the user to check the password, or try again later, or ...).

openwarrior:
@ming13 please don't make me dance. :smile: 
I really got fond of you.

openwarrior:
@danielpassos Please review the code

src/main/java/org/hawkular/client/android/auth/SecretStoreAuthzModule.java
openwarrior: @jpkrohling Problem arising here is we can't or actually don't want to do much about catch exception. It is just to be sent back via callback and handle there By showing toast or snackbar to the user which corresponding action he should take.

Even If I do something here with the catch It finally have to send whole of exception back to callback and again to be handled by callback onFailure(Exception e)

openwarrior:
@ming13 @pilhuhn @jpkrohling  If code look satisfactory please accept the pull request so that I can move ahead with more contribution. 

Generic Exceptions are removed.

src/main/java/org/hawkular/client/android/auth/SecretStoreAuthzModule.java
```diff
@@ -0,0 +1,186 @@
...
+                              final String username, final String password, final URL url) {
+
+        AsyncTask<Void, Void, HeaderAndBody> task = new AsyncTask<Void, Void, HeaderAndBody>() {
+            private Exception exception;
+
+            @Override
+            protected HeaderAndBody doInBackground(Void... params) {
+                HeaderAndBody result;
+                try {
+                    HttpProvider provider =
```
pilhuhn, line 89: Just curious - why do we here use a HttpProvider and not one of the pipes? (bear with me if I am confusing things)

src/main/java/org/hawkular/client/android/auth/SecretStoreAuthzModule.java
```diff
@@ -0,0 +1,186 @@
...
+                            new HttpRestProvider(new URL(url.toString()
+                                    + AuthData.Endpoints.ACCESS));
+                    provider.setDefaultHeader("Accept", "application/json");
+                    provider.setDefaultHeader("Authorization", "Basic "
+                            + buildLoginData(username, password));
+                    result = provider.post("");
+                    Timber.d("Called with POST to: ", new String(provider.getUrl().toString()));
```
pilhuhn: One Timber.d call is enough to show url and result (?)

pilhuhn:
A question and a comment - otherwise LGTM :+1:

src/main/java/org/hawkular/client/android/auth/SecretStoreAuthzModule.java
openwarrior, line 89: Didn't wanted to mess with the code took from cookbook of aerogear.
But can if this one look bad

src/main/java/org/hawkular/client/android/auth/SecretStoreAuthzModule.java
openwarrior: Thats right. Should i remove it with rebase or with next pull reqest?

pilhuhn: Doing it in this pr is probably best, as this is self-contained then.

src/main/java/org/hawkular/client/android/auth/SecretStoreAuthzModule.java
openwarrior, line 89: @pilhuhn  Can retain HttpProvider?

pilhuhn, line 89: @danielpassos Wdyt?

src/main/java/org/hawkular/client/android/auth/SecretStoreAuthzModule.java
openwarrior: @pilhuhn done

danielpassos:
Hey guys, I'd like to review that before merge. Please, give me some time. I'll try do it tomorrow morning

openwarrior:
@pilhuhn  Changed from AsyncTask To background Thread in thought of improvement. Please take a look to check its not got messed up.

danielpassos:
Hey @anuj1708 

First of all, sorry about my delay in review this. I have some question about this PR:
1. Why we need that? The app works without it?
2. What is the different between it and [HttpBasicAuthenticationModule](https://github.com/aerogear/aerogear-android-auth/blob/master/aerogear-android-auth/src/main/java/org/jboss/aerogear/android/authentication/basic/HttpBasicAuthenticationModule.java)?
3. You have created a new [AuthzModule](https://github.com/aerogear/aerogear-android-authz/blob/8e7520bec583ce1ad34ab9098334b795a7d7850f/aerogear-android-authz/src/main/java/org/jboss/aerogear/android/authorization/AuthzModule.java), but I can not see where it is/will be used

openwarrior:
@danielpassos  
No problem :)

1.1  We need at present to shift from connecting to Key Cloak for Access Token to using Secret Store (provide key secret pair in response of a POST request).
-- Even if we retain using access token still for new feature (QR for key/secret). Which have been added to Hawkular and to be added to android client.
1.2 At present app works without it. but not including this is limiting the functions

2 . Diff from  HttpBasicAuthenticationModule Cause its more about retaining the key secret pair for long time. and that too we never always needing to go through this to auth user as user may already have key secret  pair. which are just to be checked if right or not by any other means.
1. It will be fed to pipe for various REST requests to be performed by pipe. It needed one of the pipe module containing credentials. Shifting to (or adding key secret pair) resulted in its need.

plus. When I asked summers about need. he said its not exactly we can't so with existing code and said to extend AuthenticationModule. But as authz was already being used and have prettier code. So I considered using it .

Sorry for so long text. Was not getting ho to compress the points. :)

danielpassos:
Gotcha! It make me confused because basic auth is is  part of _auth_ but has you sad the project is using _authz_. Anyway, could you please add some code _using_ it, to we can see it really working?

danielpassos:
@secondsun Could you please take a look of that?

openwarrior:
@danielpassos  I have approx entire code ready but that is not good in looks at present. So don't want to send pul req with that.
Can i place that to my repository to show?

src/main/java/org/hawkular/client/android/auth/SecretStoreAuthzModule.java
```diff
@@ -0,0 +1,158 @@
...
+    public boolean isAuthorized() {
+        return true;
+    }
+
+    @Override
+    public boolean hasCredentials() {
+        return sessionStore.read(AuthData.NAME) != null;
+    }
+
+    public void requestAccess(final Callback<String> callback,
```
secondsun, line 83: You shouldn't be doing this.

You should be either a) using an authModule which has support for passing in a UN and PW or b) delegating the colleciton of the UN and PW to an intent and collecting the necessary tokens in a callback similar to how the intent based Oauth2AuthzModule works.

openwarrior:
I am needing to close this request. will work on it locally and resubmit. Thanks

PR #51, opened 2016-02-22
Removing Tenant from header and changes in Alert

openwarrior:
Aerogear accepts URI with GET in pipe, but not with PUT or POST.
So Complete URL are used for an alert ack and resolve.

New method savePipe is for providing general POST and PUT.

TENANT is no longer needed to be specified in the header

Complete treatment for the list of triggers connected to resource still in progress.

### DIFF
--- a/src/main/java/org/hawkular/client/android/backend/BackendClient.java
+++ b/src/main/java/org/hawkular/client/android/backend/BackendClient.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2015 Red Hat, Inc. and/or its affiliates
+ * Copyright 2015-2016 Red Hat, Inc. and/or its affiliates
  * and other contributors as indicated by the @author tags.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
@@ -136,14 +136,15 @@ public final class BackendClient {
 
     private void configurePipes(URL backendUrl, Persona persona) {
         URL pipeUrl = Urls.getUrl(backendUrl, BackendPipes.Paths.ROOT);
-
+        URL alertResolveUrl = Urls.getUrl(pipeUrl, BackendPipes.Paths.ALERT_RESOLVE);
+        URL alertAckUrl = Urls.getUrl(pipeUrl, BackendPipes.Paths.ALERT_ACKNOWLEDGE);
         List<PipeModule> pipeModules = Arrays.asList(
             getAuthorizationModule(),
             getPersonnelModule(persona));
 
         configurePipe(BackendPipes.Names.ALERTS, pipeUrl, pipeModules, Alert.class);
-        configurePipe(BackendPipes.Names.ALERT_ACKNOWLEDGE, pipeUrl, pipeModules, String.class);
-        configurePipe(BackendPipes.Names.ALERT_RESOLVE, pipeUrl, pipeModules, String.class);
+        configurePipe(BackendPipes.Names.ALERT_ACKNOWLEDGE, alertAckUrl, pipeModules, String.class);
+        configurePipe(BackendPipes.Names.ALERT_RESOLVE, alertResolveUrl, pipeModules, String.class);
         configurePipe(BackendPipes.Names.ENVIRONMENTS, pipeUrl, pipeModules, Environment.class);
         configurePipe(BackendPipes.Names.METRICS, pipeUrl, pipeModules, Metric.class);
         configurePipe(BackendPipes.Names.METRIC_DATA_AVAILABILITY, pipeUrl, pipeModules, MetricData.class);
@@ -211,16 +212,12 @@ public final class BackendClient {
 
     public void acknowledgeAlert(@NonNull Alert alert,
                                  @NonNull Callback<List<String>> callback) {
-        URI uri = Uris.getUri(String.format(BackendPipes.Paths.ALERT_ACKNOWLEDGE, alert.getId()));
-
-        readPipe(BackendPipes.Names.ALERT_ACKNOWLEDGE, uri, callback);
+        savePipe(BackendPipes.Names.ALERT_ACKNOWLEDGE, alert, callback);
     }
 
     public void resolveAlert(@NonNull Alert alert,
                              @NonNull Callback<List<String>> callback) {
-        URI uri = Uris.getUri(String.format(BackendPipes.Paths.ALERT_RESOLVE, alert.getId()));
-
-        readPipe(BackendPipes.Names.ALERT_RESOLVE, uri, callback);
+        savePipe(BackendPipes.Names.ALERT_RESOLVE, alert, callback);
     }
 
     public void getEnvironments(@NonNull Callback<List<Environment>> callback) {
@@ -288,6 +285,11 @@ public final class BackendClient {
     @SuppressWarnings("unchecked")
     private <T> void readPipe(String pipeName, URI uri, Callback<List<T>> callback) {
         getPipe(pipeName).read(getFilter(uri), callback);
+    }
+
+    @SuppressWarnings("unchecked")
+    private <T> void savePipe(String pipeName, Object object, Callback<List<T>> callback) {
+        getPipe(pipeName).save(object, callback);
     }
 
     private LoaderPipe getPipe(String pipeName) {
--- a/src/main/java/org/hawkular/client/android/backend/model/Alert.java
+++ b/src/main/java/org/hawkular/client/android/backend/model/Alert.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2015 Red Hat, Inc. and/or its affiliates
+ * Copyright 2015-2016 Red Hat, Inc. and/or its affiliates
  * and other contributors as indicated by the @author tags.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
@@ -19,6 +19,8 @@ package org.hawkular.client.android.backend.model;
 import java.util.ArrayList;
 import java.util.List;
 
+import org.jboss.aerogear.android.core.RecordId;
+
 import com.google.gson.annotations.SerializedName;
 
 import android.os.Parcel;
@@ -27,7 +29,8 @@ import android.support.annotation.NonNull;
 import android.support.annotation.VisibleForTesting;
 
 public final class Alert implements Parcelable {
-    @SerializedName("alertId")
+    @RecordId
+    @SerializedName("id")
     private String id;
 
     @SerializedName("ctime")
@@ -45,6 +48,10 @@ public final class Alert implements Parcelable {
 
     public String getId() {
         return id;
+    }
+
+    public void setId(String id) {
+        this.id = id;
     }
 
     public long getTimestamp() {
--- a/src/main/java/org/hawkular/client/android/fragment/AlertsFragment.java
+++ b/src/main/java/org/hawkular/client/android/fragment/AlertsFragment.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2015 Red Hat, Inc. and/or its affiliates
+ * Copyright 2015-2016 Red Hat, Inc. and/or its affiliates
  * and other contributors as indicated by the @author tags.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
@@ -207,7 +207,7 @@ public final class AlertsFragment extends Fragment implements AlertsAdapter.Aler
         List<Trigger> filteredTriggers = new ArrayList<>();
 
         for (Trigger trigger : triggers) {
-            if (trigger.getId().startsWith(resource.getId())) {
+            if (trigger.getId().endsWith(resource.getId())) {
                 filteredTriggers.add(trigger);
             }
         }
--- a/proguard/common.pro
+++ b/proguard/common.pro
@@ -1,1 +1,4 @@
 -dontobfuscate
+-keep class org.hawkular.client.android.backend.model.Alert {
+    public <methods>;
+}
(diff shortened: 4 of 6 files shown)


src/main/java/org/hawkular/client/android/backend/BackendClient.java
```diff
@@ -211,16 +212,13 @@ public void getAlerts(@NonNull Date startTime, @NonNull Date finishTime, @NonNul
 
     public void acknowledgeAlert(@NonNull Alert alert,
                                  @NonNull Callback<List<String>> callback) {
-        URI uri = Uris.getUri(String.format(BackendPipes.Paths.ALERT_ACKNOWLEDGE, alert.getId()));
-
-        readPipe(BackendPipes.Names.ALERT_ACKNOWLEDGE, uri, callback);
+        alert.setId(alert.getId());
```
jpkrohling: This sounds suspiciously wrong.

src/main/java/org/hawkular/client/android/backend/BackendClient.java
```diff
@@ -136,14 +136,15 @@ public void configureCommunication(@NonNull String host,
 
     private void configurePipes(URL backendUrl, Persona persona) {
         URL pipeUrl = Urls.getUrl(backendUrl, BackendPipes.Paths.ROOT);
-
+        URL alertReslveUrl = Urls.getUrl(pipeUrl, BackendPipes.Paths.ALERT_RESOLVE);
```
lucasponce: Minor thing (and not blocker, of course).
Suggestion: alertResolveUrl  as a name of the variable ?

src/main/java/org/hawkular/client/android/fragment/AlertsFragment.java
```diff
@@ -207,7 +207,7 @@ private void setUpAlertsTriggers(List<Trigger> triggers) {
         List<Trigger> filteredTriggers = new ArrayList<>();
 
         for (Trigger trigger : triggers) {
-            if (trigger.getId().startsWith(resource.getId())) {
+            if (trigger.getId().endsWith(resource.getId())) {
```
lucasponce, line 210: Perhaps, you would like to check the trigger with the tags instead the id (as commented on the IRC) ?

src/main/java/org/hawkular/client/android/backend/BackendClient.java
openwarrior: this one line I was forced to do. IDE is somehow eating the setter method if it is not used. And causing program to fail. Not leting me live even after setting minifyEnabled false

src/main/java/org/hawkular/client/android/backend/BackendClient.java
openwarrior: Name have been changed. Should be changed any minute.
And was needed to changed as Here we are defining property of Pipe(Aerogear) And for that we need to set to which URL this PUT will work upon.

src/main/java/org/hawkular/client/android/backend/BackendClient.java
jpkrohling: That sounds absolutely wrong. The IDE should not force you to do a code that is not needed. If there's a code optimization somewhere that is removing this, there should be an option to disable this "code eating" stuff. In any case, the solution is _not_ to pollute the code with unnecessary code :)

openwarrior: I am trying again. :)

src/main/java/org/hawkular/client/android/fragment/AlertsFragment.java
openwarrior, line 210: Actually that will force to Do a bigger part of code change. As at present code is not saving Tag part. Will do that later. In individual pull request. Here It was mainly to cure alert part.

lucasponce, line 210: Well, if you want to do that in a separate PR it's fine, just to note that probably the trigger.getId() might not follow the resource.getId() in some case as it is not strong linked.
So using triggert.getTags().get("resourceId") might be better.
But anyway, if this change creates issues you can do it later in a separate PR.
It is good to have a PR per topic and not mix them for easy review.

openwarrior, line 210: Yes. And this one got a bit messed up already ;)
and thats why I placed that line in commit "Complete treatment for the list of triggers connected to resource still in progress". As a reminder for me.

tsegismont:
LGTM with respect to tenant header removal.

src/main/java/org/hawkular/client/android/backend/BackendClient.java
openwarrior: @jpkrohling Thanks. Done 
Changed the progourd to stop looking for public methods of Alert class

src/main/java/org/hawkular/client/android/backend/BackendClient.java
```diff
@@ -290,6 +287,11 @@ public void getTriggers(@NonNull Callback<List<Trigger>> callback) {
         getPipe(pipeName).read(getFilter(uri), callback);
     }
 
+    @SuppressWarnings("unchecked")
+    private <T> void savePipe(String pipeName, Object object, Callback<List<T>> callback, Class pipeClass) {
+        getPipe(pipeName).save(pipeClass.cast(object), callback);
```
danielpassos: I think we don't need `pipeClass` as a method parameter and also `pipeClass.cast(object)`

Any special reason to use that? You had some problem without it?

openwarrior: @danielpassos Thanks. I didn't checked. Just thought it will be needing to know from which class object relate to. But checked its not. :)

openwarrior:
@ming13 please review :)

src/main/java/org/hawkular/client/android/backend/BackendClient.java
```diff
@@ -290,6 +287,11 @@ public void getTriggers(@NonNull Callback<List<Trigger>> callback) {
         getPipe(pipeName).read(getFilter(uri), callback);
     }
 
+    @SuppressWarnings("unchecked")
+    private <T> void savePipe(String pipeName, Object object, Callback<List<T>> callback) {
```
arturdryomov, line 291: Even if original `Pipe` interface uses `save` keyword, maybe `write` be more suitable. Also, and more important, you can change signature to `savePipe(String pipeName, T pipeItem, Callback<List<T>> callback)`.

src/main/java/org/hawkular/client/android/backend/BackendClient.java
```diff
@@ -136,14 +136,15 @@ public void configureCommunication(@NonNull String host,
...
+        URL alertResolveUrl = Urls.getUrl(pipeUrl, BackendPipes.Paths.ALERT_RESOLVE);
+        URL alertAckUrl = Urls.getUrl(pipeUrl, BackendPipes.Paths.ALERT_ACKNOWLEDGE);
         List<PipeModule> pipeModules = Arrays.asList(
             getAuthorizationModule(),
             getPersonnelModule(persona));
 
         configurePipe(BackendPipes.Names.ALERTS, pipeUrl, pipeModules, Alert.class);
-        configurePipe(BackendPipes.Names.ALERT_ACKNOWLEDGE, pipeUrl, pipeModules, String.class);
-        configurePipe(BackendPipes.Names.ALERT_RESOLVE, pipeUrl, pipeModules, String.class);
+        configurePipe(BackendPipes.Names.ALERT_ACKNOWLEDGE, alertAckUrl, pipeModules, String.class);
```
arturdryomov, line 146: You previous change lead to a little confusion here my friend. `a34f0421` didn’t consider that we have a single `ROOT` path. So, you should do this instead:

``` java
// BackendClient.java

configurePipe(BackendPipes.Names.ALERT_ACKNOWLEDGE, pipeUrl, pipeModules, String.class);
configurePipe(BackendPipes.Names.ALERT_RESOLVE, pipeUrl, pipeModules, String.class);
...
configurePipe(BackendPipes.Names.PERSONA, pipeUrl, pipeModules, Persona.class);
configurePipe(BackendPipes.Names.PERSONAS, pipeUrl, pipeModules, Persona.class);
```

``` java
// BackendPipes.java

public static final String PERSONA = "accounts/personas/current";
public static final String PERSONAS = "accounts/personas";
```

And then you can remove your introduced variables. This should be much better.

src/main/java/org/hawkular/client/android/backend/model/Alert.java
```diff
@@ -47,6 +50,10 @@ public String getId() {
         return id;
     }
 
+    public void setId(String id) {
```
arturdryomov, line 53: Is not used anywhere.

src/main/java/org/hawkular/client/android/backend/model/Alert.java
```diff
@@ -27,7 +29,8 @@
 import android.support.annotation.VisibleForTesting;
 
 public final class Alert implements Parcelable {
-    @SerializedName("alertId")
+    @RecordId
```
arturdryomov, line 32: Is not used anywhere.

openwarrior, line 32: I was using save() and that needed this. cause it needed a object with RecordId not null to place a PUT request.
Lemme check what is there for write() :)

src/main/java/org/hawkular/client/android/backend/model/Alert.java
openwarrior, line 53: setId and getId are getter and setter which you are bound to make with @RecordId thing

src/main/java/org/hawkular/client/android/backend/BackendClient.java
openwarrior, line 291: Using modified Pipe will be quite troublesome I think.
And using T giving error with calling as "Found org.hawkular.client.android.backend.model.Alert needed Java.lang.String

src/main/java/org/hawkular/client/android/backend/model/Alert.java
arturdryomov, line 32: Oh. That’s a bummer.

src/main/java/org/hawkular/client/android/backend/BackendClient.java
openwarrior, line 146: Actually got bounded a bit cause of save again. Can't pass a URI to it. So bound to give it whole of URL in starting only. Asking Summers for help but I think he didn't had time today. But example link he gave was following this stuff.

src/main/java/org/hawkular/client/android/backend/BackendClient.java
arturdryomov, line 291: Are you sure that you’ve swapped `Object` with `T`?

src/main/java/org/hawkular/client/android/backend/model/Alert.java
openwarrior, line 32: Yeah. Faced a lot of difficulty cause progourd was wiping the method and program was not working without it.
Its how it is made to be.
Alert ACK is now with a PUT request

src/main/java/org/hawkular/client/android/backend/BackendClient.java
arturdryomov, line 146: I don’t see any relation to `save` at this point. You are configuring pipes.

src/main/java/org/hawkular/client/android/backend/BackendClient.java
openwarrior, line 291: yes I am. I was shocked too. But Object was working good so just moved on.
Even retried after you said

src/main/java/org/hawkular/client/android/backend/BackendClient.java
openwarrior, line 146: yeah. That means setting URL. To which i cant later add a URI to complete the address cause save take object and callback. And not URI. Quite diff from GET request method it provide

arturdryomov, line 146: Anyhow, you can change personas handling at least. For posting things you can use a slightly different approach. Use `backendUrl` and specify full path at `BackendPaths`, i. e. with the `hawkular` prefix. Will it work?

openwarrior, line 146: I am really sorry but not getting your point

arturdryomov, line 146:
``` java
// BackendClient.java

configurePipe(BackendPipes.Names.ALERT_ACKNOWLEDGE, backendUrl, pipeModules, String.class);
configurePipe(BackendPipes.Names.ALERT_RESOLVE, backendUrl, pipeModules, String.class);
```

``` java
// BackendPipes.java

public static final String ALERT_ACKNOWLEDGE = "hawkular/alerts/ack";
public static final String ALERT_RESOLVE = "hawkular/alerts/resolve";
```

openwarrior, line 146: Didn't alter url for persona in fear of hearing it be useless change. cause will have no extra effect.
Should I change them to `backendUrl`

pilhuhn:
Thanks @ming13 , @tsegismont , @danielpassos , @jpkrohling for reviewing :clap:

pilhuhn:
Thank you @anuj1708

openwarrior:
@pilhuhn Your always welcome Sir

PR #52, opened 2016-02-23
Change in trigger identification for resources

openwarrior:
Trigger ID can now be independent of resource ID.
So using the tag to identify it.

### DIFF
--- a/src/androidTest/java/org/hawkular/client/android/backend/model/TriggerTester.java
+++ b/src/androidTest/java/org/hawkular/client/android/backend/model/TriggerTester.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2015 Red Hat, Inc. and/or its affiliates
+ * Copyright 2015-2016 Red Hat, Inc. and/or its affiliates
  * and other contributors as indicated by the @author tags.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
@@ -16,6 +16,9 @@
  */
 package org.hawkular.client.android.backend.model;
 
+import java.util.HashMap;
+import java.util.Map;
+
 import org.assertj.core.api.Assertions;
 import org.hawkular.client.android.util.Parceler;
 import org.hawkular.client.android.util.Randomizer;
@@ -28,7 +31,9 @@ import android.support.test.runner.AndroidJUnit4;
 public final class TriggerTester {
     @Test
     public void parcelable() {
-        Trigger originalTrigger = new Trigger(Randomizer.generateString());
+        Map<String, String> test = new HashMap<>();
+        test.put(Randomizer.generateString(), Randomizer.generateString());
+        Trigger originalTrigger = new Trigger(Randomizer.generateString(), test);
         Trigger parceledTrigger = Parceler.parcel(Trigger.CREATOR, originalTrigger);
 
         Assertions.assertThat(parceledTrigger.getId()).isEqualTo(originalTrigger.getId());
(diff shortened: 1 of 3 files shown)


lucasponce:
PR looks good to me
Perhaps one side question is why don't use the hawkular-alerts-api package as base of the org.hawkular.client.android.backend.model package.
There you have the tools for dealing with the json -> object serialization of alerts and it could save some of the work done here.
Perhaps you can extend them in a class where you can implement the Parcel interface if needed.

openwarrior:
@lucasponce I will surely look at that :)

src/androidTest/java/org/hawkular/client/android/backend/model/TriggerTester.java
```diff
@@ -28,7 +31,9 @@
 public final class TriggerTester {
     @Test
     public void parcelable() {
-        Trigger originalTrigger = new Trigger(Randomizer.generateString());
+        Map<String, String> test =new HashMap<>();
```
jpkrohling: Minor detail: add a space after "="

src/main/java/org/hawkular/client/android/fragment/AlertsFragment.java
```diff
@@ -207,9 +207,10 @@ private void setUpAlertsTriggers(List<Trigger> triggers) {
         List<Trigger> filteredTriggers = new ArrayList<>();
 
         for (Trigger trigger : triggers) {
-            if (trigger.getId().endsWith(resource.getId())) {
-                filteredTriggers.add(trigger);
-            }
+                if ( trigger.getTags()!=null && trigger.getTags().containsKey("resourceId") && trigger.getTags().get
```
jpkrohling: The indentation here looks odd.

openwarrior: Cure is on the way :)

openwarrior: @jpkrohling Anything else sir?

jpkrohling: Spacing around `!=` on `getTags()!=null` :)

jpkrohling: Also, it would be better to make it more readable, by splitting the conditions each on one line:

```
trigger.getTags()!=null 
&& trigger.getTags().containsKey("resourceId") 
&& trigger.getTags().get("resourceId").equals(resource.getId())
```

In fact, I believe this could be simplified:

```
trigger.getTags()!=null 
&& resource.getId().equals(trigger.getTags().get("resourceId"))
```

As long as `resource` is never null.

openwarrior: there may be tag or no tag and that tag may contain or may not contain resourceId so I thought to be on safe side

jpkrohling: The first part you are dealing on the `!= null`. The second part (may or not contain resourceId) is dealt with the equals on the `resource.getId()`: if `getTags().get("resourceId")` returns null, it won't be equal to `resource.getId()`.

openwarrior: got your point. that can be done

pilhuhn:
Thanks @anuj1708 :clap:

PR #54, opened 2016-03-01
Secret Store Authorization Implementation

openwarrior:
SecretStoreAuthzModule:-
Performs all related work of generating new key and secret pair, Storing them as account,
Retrieving stored account, Deleting stored account

ModuleKeeper:-
This class have sole reason for storing the Authentication module.
So that can be accessed from anywhere in the code at any time.

Session:-
Body to store key, secret and other related stuff

AuthData:-
Contain Strings which are to be used by package.

### DIFF
--- a/src/main/java/org/hawkular/client/android/auth/AuthData.java
+++ b/src/main/java/org/hawkular/client/android/auth/AuthData.java
@@ -1,0 +1,47 @@
+/*
+ * Copyright 2015-2016 Red Hat, Inc. and/or its affiliates
+ * and other contributors as indicated by the @author tags.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *    http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.hawkular.client.android.auth;
+
+/**
+ * Contain Hard Coded Strings which are to be used by package.
+ */
+
+public final class AuthData {
+    public static final String NAME = "hawkular";
+    public static final String STORE = "sessionStore";
+
+    public static final class Endpoints {
+        private Endpoints() {
+        }
+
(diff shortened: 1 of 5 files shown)


src/main/java/org/hawkular/client/android/auth/SecretStoreAuthzModule.java
```diff
@@ -0,0 +1,177 @@
...
+    @Override
+    public boolean handleError(HttpException exception) {
+        return false;
+    }
+
+    private void addAccount(String data) throws JSONException {
+        JSONObject auth = new JSONObject(data);
+        OAuth2AuthzSession session = new OAuth2AuthzSession();
+        session.setAccountId(AuthData.NAME);
+        session.setAccessToken(auth.getString("key") + ":" + auth.getString("secret"));
```
jpkrohling: Is this how it's supposed to be used? Secret Store's tokens are not OAuth Tokens.

openwarrior: @jpkrohling  I was just using that to store key:secret . But right, Its not a good practice. repairing it

openwarrior: @jpkrohling Changed to customized Session class. :)

openwarrior:
@danielpassos  Please review the pull request. And also ping summers as I can't ping him. Thanks

src/main/java/org/hawkular/client/android/auth/AuthData.java
```diff
@@ -0,0 +1,43 @@
...
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.hawkular.client.android.auth;
+
+public final class AuthData {
```
pilhuhn, line 23: Can you please add a class comment about that this class is supposed to do?

Could those constants be moved to `auth.Session` class?

pilhuhn:
So that is basically a re-make of the other PR that you did and then closed to be able to do other work first?

openwarrior:
@pilhuhn Approx. With some little enhancements told by summers and jpkrohling

src/main/java/org/hawkular/client/android/auth/AuthData.java
openwarrior, line 23: Adding comment to this class but can't be moved as session is something to be stored in aerogear store but this class will be needed multiple times as a map for auth module :)

PR #55, opened 2016-03-08
Replace WebView login by local Secret Strore login

openwarrior:
Login Activity is Introduced.
A large chunk of code have be taken from end of `AuthorizationActivity` and
placed in `LoginActivity`. As functionality is divided.

New Secret Store Authz module is being used for authorization.

`expiresAt:null` in token is handled.


pilhuhn:
Thank you :+1:

openwarrior:
thank you

PR #57, opened 2016-04-06
Initial work on saving and visiting tokens

openwarrior:
Token : A model for token to be saved
TokensAdapter : Transforms a list of Tokens to a human-readable interpretation.
TokensFragment : To be added in container to present data.

'White Plus' icons to be added on Floating Action Bar


src/main/java/org/hawkular/client/android/backend/model/Token.java
```diff
@@ -0,0 +1,85 @@
...
+        @Override public Token createFromParcel(Parcel parcel) {
+            return new Token(parcel);
+        }
+
+        @Override
+        public Token[] newArray(int size) {
+            return new Token[size];
+        }
+    };
+    @RecordId
```
pilhuhn, line 38: Leave an empty line before this

pilhuhn:
Thanks @anuj1708

src/main/java/org/hawkular/client/android/backend/model/Token.java
openwarrior, line 38: Added in coming PR :)

openwarrior:
Thanks for your support @pilhuhn

PR #60, opened 2016-04-27
 Available personas in navigation drawer

openwarrior:
At present resources are to be refreshed after changing the persona is changed.
No token is saved by default. That will be changed to save to first token itself.


openwarrior:
New changes in Hawkular server (Removing support for tokens) do not need this code. So canceling the PR. :)

PR #61, opened 2016-05-25
Change in model of Alert and Trigger added Note

openwarrior:
Alert - Add extra class Lister to turn Alert properly parcelable, few extra items for additional features

Trigger - add detail field

Note - Model for Comments on Alert

### DIFF
--- a/src/androidTest/java/org/hawkular/client/android/backend/model/NoteTester.java
+++ b/src/androidTest/java/org/hawkular/client/android/backend/model/NoteTester.java
@@ -16,8 +16,6 @@
  */
 package org.hawkular.client.android.backend.model;
 
-import java.util.ArrayList;
-
 import org.assertj.core.api.Assertions;
 import org.hawkular.client.android.util.Parceler;
 import org.hawkular.client.android.util.Randomizer;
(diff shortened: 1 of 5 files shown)


src/androidTest/java/org/hawkular/client/android/backend/model/NoteTester.java
```diff
@@ -0,0 +1,40 @@
...
+    @Test
+    public void parcelable() {
+        Note originalNote = new Note(
+                Randomizer.generateString(), Randomizer.generateNumber(), Randomizer.generateString());
+
+        Note parceledNote = Parceler.parcel(Note.CREATOR, originalNote);
+
+        Assertions.assertThat(parceledNote.getUser()).isEqualTo(originalNote.getUser());
+        Assertions.assertThat(parceledNote.getMessage()).isEqualTo(originalNote.getMessage());
+        Assertions.assertThat(parceledNote.getTimestamp()).isEqualTo(originalNote.getTimestamp());
```
lucasponce, line 38: In Hawkular Alerts is Note.getCtime(), just in case you want to align it with the backend model.

lucasponce:
Hi Anuj.
I'm not familiar with specific of android classes but I didn't see anything suspicious.

src/androidTest/java/org/hawkular/client/android/backend/model/NoteTester.java
openwarrior, line 38: ok thanks :)

PR #63, opened 2016-06-09
Update Support Libs, Aerogear libs, &  ButterKnife

openwarrior:


src/main/java/org/hawkular/client/android/activity/DrawerActivity.java
```diff
@@ -72,29 +70,24 @@
...
-    @Bind(R.id.text_persona)
     TextView persona;
 
-    @Bind(R.id.list_personas)
     ListView personas;
 
-    @Bind(R.id.layout_personas)
     ViewGroup personasLayout;
 
-    @Bind(R.id.image_personas_action)
```
danielpassos, line 97: Why `@Bind` was replaced of `ButterKnife.findById`?

openwarrior, line 97: With new support lib. items in header of nav drawers are not directly visible.

danielpassos:
@anuj1708 You need to update the `.travis.yml` with yours new updates to make @travis-ci happy :)

https://github.com/hawkular/hawkular-android-client/blob/master/.travis.yml#L8-L9

openwarrior:
@danielpassos travis is happy now :smile:

.travis.yml
```diff
@@ -5,8 +5,9 @@ jdk:
 
 android:
   components:
-    - android-22
-    - build-tools-22.0.1
+    - tools
```
pilhuhn, line 8: What is 'tools' ?

openwarrior, line 8: There is an issue in travis cause of which it was saying that i cant not find build-tools-23.0.2. I found on net and found that it is gone if you add tools, So I don't know much at present except of travis fail to find build-tools-23.0.2 without it.

openwarrior, line 8: https://github.com/travis-ci/travis-ci/issues/5036#issuecomment-153297292

here is from where I took this :)

pilhuhn:
Thanks @anuj1708

openwarrior:
Thanks :)

PR #65, opened 2016-07-04
Connect explorer to metrics in chart view

openwarrior:
Change fetch type from data to bucket
Add Counter metric implementation
Place Time range selector menu in chart view


src/main/java/org/hawkular/client/android/backend/BackendClient.java
```diff
@@ -255,26 +258,45 @@ public void getMetrics(@NonNull Environment environment, @NonNull Resource resou
...
 
-        URI uri = Uris.getUri(String.format(BackendPipes.Paths.METRIC_DATA_AVAILABILITY, resource.getId()), parameters);
+        URI uri = Uris.getUri(String.format(BackendPipes.Paths.METRIC_DATA_AVAILABILITY,
+                Uris.getEncodedParameter(metric.getId())), parameters);
 
         readPipe(BackendPipes.Names.METRIC_DATA_AVAILABILITY, uri, callback);
     }
 
-    public void getMetricDataGauge(@NonNull Metric metric,
+    public void getMetricDataCounter(@NonNull Metric metric, long bucket,
```
pilhuhn: I think you should try to not duplicate this via copy&paste, but have one more generic method `getMetricData` where you pass in the type as parameter. It is a bit more tricky for availability above, but there you could at least extract the setting of start/end/buckets into a helper method

src/main/java/org/hawkular/client/android/backend/model/MetricAvailabilityBucket.java
```diff
@@ -0,0 +1,76 @@
...
+ * limitations under the License.
+ */
+package org.hawkular.client.android.backend.model;
+
+import com.google.gson.annotations.SerializedName;
+
+import android.os.Parcel;
+import android.os.Parcelable;
+
+public class MetricAvailabilityBucket implements Parcelable {
```
pilhuhn: Isn't that a sub-class of MetricBucket below?

src/main/java/org/hawkular/client/android/fragment/MetricAvailabilityFragment.java
```diff
@@ -71,14 +76,21 @@ private Defaults() {
...
+
     @BindView(R.id.chart)
     ColumnChartView chart;
 
     @BindView(R.id.content)
     SwipeRefreshLayout contentLayout;
 
     @State
-    ArrayList<MetricData> metricData;
+    ArrayList<MetricAvailabilityBucket> metricBucket;
```
pilhuhn: It is good practice to use an interface on the declaration if possible.

src/main/java/org/hawkular/client/android/fragment/MetricAvailabilityFragment.java
```diff
@@ -172,20 +252,28 @@ private void setUpChartColumns() {
...
 
     private List<Column> getChartColumns() {
-        List<Column> chartColumns = new ArrayList<>(metricData.size());
+        List<Column> chartColumns = new ArrayList<>(metricBucket.size());
+
+        for (MetricAvailabilityBucket metricBucket : this.metricBucket) {
+            MetricAvailability metricAvailability = null;
+            if (metricBucket.getValue().equals("NaN")) {
+                metricAvailability = MetricAvailability.from("unknown");
+            } else if (Float.parseFloat(metricBucket.getValue()) >= .5) {
```
pilhuhn, line 271: There is now since hawkular-services 0.0.5 also a value "admin" available.
You should probably test this whole section against 0.0.5

src/main/java/org/hawkular/client/android/backend/BackendClient.java
```diff
@@ -131,8 +132,10 @@ private void configurePipes(URL backendUrl, Persona persona) {
         configurePipe(BackendPipes.Names.FEED_REC_RESOURCES, pipeUrl, pipeModules, Resource.class);
```
pilhuhn, line 135: In hawkular-services 0.0.5 the inventory api has been moved to `/deprecated`.  you may need to change the pipes.
There is a new api under the endpoint that you know, but which is not compatible. So I guess for the moment using `http://<host>:<port>/hawkular/inventory/deprecated` as base is best.

You could also check for inventory version like this is done in the ruby-client: https://github.com/hawkular/hawkular-client-ruby/commit/7a857ac051ea2b3d81ddd690721cdb2382222012

openwarrior, line 135: Setting up hawkular-services 0.0.5 . Till now was working with alpha 11 :)

src/main/java/org/hawkular/client/android/backend/model/MetricAvailabilityBucket.java
openwarrior: Actually no. I chose bad naming but 
Gauge and counter bucket send data in same fashion but Availability bucket is completely different. So class named MetricAvailabilityBucket only for Availabilility and MetricBucket for Counter and Gauge Buckets

PR #66, opened 2016-07-11
switch to hawkular-service 0.0.5

openwarrior:
Auth system and URIs are updated.

### DIFF
--- a/src/main/java/org/hawkular/client/android/auth/AuthData.java
+++ b/src/main/java/org/hawkular/client/android/auth/AuthData.java
@@ -23,14 +23,12 @@ package org.hawkular.client.android.auth;
 public final class AuthData {
     public static final String NAME = "hawkular";
     public static final String STORE = "sessionStore";
-    public static final String DESCRIPTION = "{\"attributes\":{\"description\":\"" +
-            "'Token created via Hawkular Android Client'\"}}";
 
     public static final class Endpoints {
         private Endpoints() {
         }
 
-        public static final String ACCESS = "/secret-store/v1/tokens/create";
+        public static final String ACCESS = "/hawkular/metrics/tenants";
         public static final String PERSONA = "/hawkular/accounts/personas/current";
     }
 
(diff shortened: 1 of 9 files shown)


src/main/java/org/hawkular/client/android/auth/AuthData.java
```diff
@@ -23,14 +23,12 @@
...
     public static final String STORE = "sessionStore";
-    public static final String DESCRIPTION = "{\"attributes\":{\"description\":\"" +
-            "'Token created via Hawkular Android Client'\"}}";
 
     public static final class Endpoints {
         private Endpoints() {
         }
 
-        public static final String ACCESS = "/secret-store/v1/tokens/create";
+        public static final String ACCESS = "/hawkular/metrics/tenants";
```
pilhuhn, line 31: What is this for?

src/main/java/org/hawkular/client/android/backend/BackendPipes.java
```diff
@@ -59,18 +59,18 @@ private Paths() {
...
+        public static final String FEED_RESOURCES = "inventory/deprecated/feeds/%s/resources";
+        public static final String FEED_REC_RESOURCES = "inventory/deprecated/feeds/%s/resources/%s/recursiveChildren";
+        public static final String METRICS = "inventory/deprecated/%s/resources/%s/metrics";
         public static final String METRIC_DATA_AVAILABILITY= "metrics/availability/%s/data";
         public static final String METRIC_DATA_COUNTER = "metrics/counters/%s/data";
         public static final String METRIC_DATA_GAUGE = "metrics/gauges/%s/data";
-        public static final String PERSONA = "hawkular/accounts/personas/current";
-        public static final String PERSONAS = "hawkular/accounts/personas";
-        public static final String RESOURCES = "inventory/%s/resources";
+        public static final String PERSONA = "hawkular/metrics/tenants";
```
pilhuhn, line 71: What are those for?

src/main/java/org/hawkular/client/android/auth/AuthData.java
openwarrior, line 31: When person ender username and password. I have to check if it is correct or not. Earlier it was we could check on the persona but as it is not there I found this uri as a good point to verify using HTTP code of response

src/main/java/org/hawkular/client/android/backend/BackendPipes.java
openwarrior, line 71: Earlier Persona use to give tenant-id. But as tenant is not totally replaced by persona so i am just using it so contain tenant. To make the code moving and make some minor changes according to what he says

PR #67, opened 2016-07-13
Bump libs

danielpassos:
This PR:
1. Remove deprecated SDK manager  
2. Update Gradle tools version to enable [Instant Run](https://developer.android.com/studio/run/index.html#instant-run)
3. Update project to Android M (API level 24)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -6,8 +6,8 @@ jdk:
 android:
   components:
     - tools
-    - android-23
-    - build-tools-23.0.2
+    - android-24
+    - build-tools-24.0.0
     - extra-android-m2repository
     - extra-google-m2repository
 
--- a/build.gradle
+++ b/build.gradle
@@ -29,8 +29,8 @@ apply plugin: "com.android.application"
 apply plugin: "android-apt"
 
 android {
-    compileSdkVersion 23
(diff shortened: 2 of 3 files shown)


danielpassos:
@anuj1708 Can you make sure nothing broke?

openwarrior:
@danielpassos Rest is all fine but `Instant Run: Proguard is not compatible with instant run. It has been disabled for debug` . But number of libraries used in project is so high that if progaurd is off then it will need multidexing. In that case also instant run will not work.

We are in situation where we can not use Instant Run anyway possible.

To run this PR. Instant Run is needed to be turned off otherwise progaurd is turned off for debug and hence method ref exceed max value for single dex. Steps to disable instant run are :-
1.     Open the Settings or Preferences dialog: On Windows or Linux, select File > Settings from the main menu. On Mac OSX, select Android Studio > Preferences from the main menu.
2.     Navigate to Build, Execution, Deployment > Instant Run.
3.     Uncheck the box next to Enable Instant run

danielpassos:
@anuj1708 If nothing was broke, let's merge it and I'll work later in get Instant run working.

openwarrior:
@danielpassos Sure :+1:

PR #69, opened 2016-07-25
Manage empty metric buckets

openwarrior:


src/main/java/org/hawkular/client/android/backend/model/MetricBucket.java
```diff
@@ -26,6 +26,9 @@
     @SerializedName("end")
     protected long endTimestamp;
 
+    @SerializedName("empty")
+    protected String empty;
```
pilhuhn: This very much sounds like it should be a boolean ?

openwarrior: Yeah but while saving it in serialization, that convert to 0/1 and cause pain. So i thought it would be better if it remain String only

src/main/java/org/hawkular/client/android/backend/model/MetricAvailabilityBucket.java
```diff
@@ -39,6 +39,7 @@ public MetricAvailabilityBucket createFromParcel(Parcel in) {
 
     protected MetricAvailabilityBucket(Parcel in) {
         value = in.readString();
+        empty = in.readString().equals("true");
```
openwarrior, line 42: @pilhuhn I feel it fills that gap from both end. `android.os.Parcel` do not have real and write method for boolean so forced to save in string or any other form.

PR #70, opened 2016-07-25
Place favorite metrics to home screen for better user experience

openwarrior:
- Move favorite metrics to home screen
- Add CanonicalPath utility


danielpassos:
👍

PR #73, opened 2016-07-26
Fix uncheck menu item

danielpassos:
Today when you click in a link on menu and again in another one, the previous still checked/highlight.

I also added midp icons for the drawer menu and explore icon missed


openwarrior:
All good :+1:

PR #74, opened 2016-07-27
Remove token and related stuff

openwarrior:
Since Keyclock is removed from Hawkular, tokens are no longer supported in hawkular


danielpassos:
👍

PR #79, opened 2016-07-29
Switch app from Fragment to support Fragment

openwarrior:
Support Fragment provide persistent behaviour and also help to use other tools from support libs.


danielpassos:
👍

PR #81, opened 2016-08-08
Implement websocket to perform Operation over resources

openwarrior:
- Add properties and params to Operation model.
- Add utility classes to perform websocket connection related tasks.
-  SSL compatiblity is not working yet.
- Operations and custom params on operations are completely working.

### DIFF
--- a/mobile/src/main/java/org/hawkular/client/android/fragment/ConfirmOperationFragment.java
+++ b/mobile/src/main/java/org/hawkular/client/android/fragment/ConfirmOperationFragment.java
@@ -29,7 +29,6 @@ import org.hawkular.client.android.backend.model.Resource;
 import org.hawkular.client.android.util.Fragments;
 import org.hawkular.client.android.util.OperationManager;
 import org.jboss.aerogear.android.core.Callback;
-import org.json.JSONArray;
 import org.json.JSONException;
 import org.json.JSONObject;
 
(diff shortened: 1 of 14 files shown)


mobile/src/main/java/org/hawkular/client/android/fragment/ConfirmOperationFragment.java
```diff
@@ -0,0 +1,255 @@
...
+
+                    case "float": {
+                        row = (TableRow) LayoutInflater.from(getActivity()).inflate(R.layout.row_edit, null);
+                        EditText data = (EditText) row.findViewById(R.id.data);
+                        data.setText(operationParameter.getDefaultValue());
+                        data.setInputType(InputType.TYPE_CLASS_NUMBER | InputType.TYPE_NUMBER_FLAG_DECIMAL);
+                        break;
+                    }
+
+                    default: {
```
pilhuhn: I guess this is meant for boolean. Better have this explicit as `case "boolean":` and use the default for cases where the incoming data is not known. As is it works, but it is not obvious when you read the code.

openwarrior:
on it

pilhuhn:
Thanks @anuj1708 that is great.

openwarrior:
@pilhuhn Thanks

PR #82, opened 2016-08-09
Remove depecrated api calls

openwarrior:
- add path to models that do not have already.
- change api calls to new api

### DIFF
--- a/mobile/src/main/java/org/hawkular/client/android/backend/BackendClient.java
+++ b/mobile/src/main/java/org/hawkular/client/android/backend/BackendClient.java
@@ -134,7 +134,7 @@ public final class BackendClient {
         configurePipe(BackendPipes.Names.ENVIRONMENTS, pipeUrl, pipeModules, Environment.class);
         configurePipe(BackendPipes.Names.FEEDS, pipeUrl, pipeModules, Feed.class);
         configurePipe(BackendPipes.Names.FEED_METRICS, pipeUrl, pipeModules, Metric.class);
-        configurePipe(BackendPipes.Names.FEED_REC_RESOURCES, pipeUrl, pipeModules, Resource.class);
+        configurePipe(BackendPipes.Names.FEED_CHILD_RESOURCES, pipeUrl, pipeModules, Resource.class);
         configurePipe(BackendPipes.Names.FEED_RESOURCES, pipeUrl, pipeModules, Resource.class);
         configurePipe(BackendPipes.Names.METRICS, pipeUrl, pipeModules, Metric.class);
         configurePipe(BackendPipes.Names.METRIC_DATA_AVAILABILITY, pipeUrl, pipeModules,
@@ -234,16 +234,14 @@ public final class BackendClient {
     }
 
     public void getOpreations(@NonNull Callback<List<Operation>> callback, Resource resource) {
-        URI uri = Uris.getUri(String.format(BackendPipes.Paths.OPERATIONS,
-                CanonicalPath.getByString(resource.getPath()).getFeed(),
-                Uris.getEncodedParameter(resource.getType().getId())));
+        URI uri = Uris.getUri(CanonicalPath.getByString(resource.getType().getPath())
+                .fix(BackendPipes.Paths.OPERATIONS));
 
         readPipe(BackendPipes.Names.OPERATIONS, uri, callback);
     }
 
     public void getResourcesFromFeed(@NonNull Callback<List<Resource>> callback, Feed feed) {
-        URI uri = Uris.getUri(String.format(BackendPipes.Paths.FEED_RESOURCES,
-                Uris.getEncodedParameter(feed.getId())));
+        URI uri = Uris.getUri(CanonicalPath.getByString(feed.getPath()).fix(BackendPipes.Paths.FEED_RESOURCES));
 
         readPipe(BackendPipes.Names.FEED_RESOURCES, uri, callback);
     }
@@ -251,16 +249,16 @@ public final class BackendClient {
 
     public void getRecResourcesFromFeed(@NonNull Callback<List<Resource>> callback, Resource resource) {
 
-            URI uri = Uris.getUri(String.format(BackendPipes.Paths.FEED_REC_RESOURCES,resource.getFeed(),
-                    CanonicalPath.getByString(resource.getPath()).getResource()));
+            URI uri = Uris.getUri(CanonicalPath.getByString(resource.getPath()).
+                    fix(BackendPipes.Paths.FEED_CHILD_RESOURCES));
 
-        readPipe(BackendPipes.Names.FEED_REC_RESOURCES, uri, callback);
+        readPipe(BackendPipes.Names.FEED_CHILD_RESOURCES, uri, callback);
     }
 
 
     public void getMetricsFromFeed(@NonNull Callback<List<Metric>> callback, Resource resource) {
-        URI uri = Uris.getUri(String.format(BackendPipes.Paths.FEED_METRICS, resource.getFeed(),
-                CanonicalPath.getByString(resource.getPath()).getResource()));
+        URI uri = Uris.getUri(CanonicalPath.getByString(resource.getPath()).
+                fix(BackendPipes.Paths.FEED_METRICS));
 
         readPipe(BackendPipes.Names.FEED_METRICS, uri, callback);
     }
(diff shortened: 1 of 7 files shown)


mobile/src/main/java/org/hawkular/client/android/util/CanonicalPath.java
```diff
@@ -20,6 +20,8 @@
 import java.util.HashMap;
 import java.util.LinkedList;
 import java.util.List;
+import java.util.regex.Matcher;
```
pilhuhn: Where is that needed?

mobile/src/main/java/org/hawkular/client/android/backend/BackendPipes.java
```diff
@@ -60,19 +60,19 @@ private Paths() {
         public static final String ALERT_ACKNOWLEDGE = "alerts/ack";
         public static final String ALERT_NOTE = "alerts/note";
         public static final String ALERT_RESOLVE = "alerts/resolve";
-        public static final String ENVIRONMENTS = "inventory/deprecated/environments";
-        public static final String FEEDS = "inventory/deprecated/feeds";
-        public static final String FEED_METRICS = "inventory/deprecated/feeds/%s/resources/%s/metrics";
-        public static final String FEED_RESOURCES = "inventory/deprecated/feeds/%s/resources";
-        public static final String FEED_REC_RESOURCES = "inventory/deprecated/feeds/%s/resources/%s/children";
-        public static final String METRICS = "inventory/deprecated/%s/resources/%s/metrics";
+        public static final String ENVIRONMENTS = "inventory/traversal/type=e";
```
pilhuhn, line 63: @jiri-kremser Could you please have a quick look at the `inventory/*` paths if they make sense for the new new api?

mobile/src/main/java/org/hawkular/client/android/util/CanonicalPath.java
openwarrior: @pilhuhn oops. I missed to remove that. I used that at first but removed later

mobile/src/main/java/org/hawkular/client/android/backend/BackendPipes.java
jkremser, line 63: feeds, feeds_resources, resources look good to me, I am not 100% sure about the others @metlos would be better for this, I haven't completelly wrapped my head around it yet

mobile/src/main/java/org/hawkular/client/android/backend/BackendPipes.java
```diff
@@ -60,19 +60,19 @@ private Paths() {
...
+        public static final String FEEDS = "inventory/traversal/type=f";
+        public static final String FEED_METRICS = "inventory/traversal/f;{f}/{r}/rl;incorporates/type=m";
+        public static final String FEED_RESOURCES = "inventory/traversal/f;{f}/type=r";
+        public static final String FEED_CHILD_RESOURCES = "inventory/traversal/f;{f}/{r}/type=r";
```
metlos: `inventory/traversal/f;{f}/{r}/rl;isParentOf/type=r` to be equivalent to what was there originally. Note that `isParentOf` is more general than `contains`

mobile/src/main/java/org/hawkular/client/android/backend/BackendPipes.java
```diff
@@ -60,19 +60,19 @@ private Paths() {
...
         public static final String ALERT_RESOLVE = "alerts/resolve";
-        public static final String ENVIRONMENTS = "inventory/deprecated/environments";
-        public static final String FEEDS = "inventory/deprecated/feeds";
-        public static final String FEED_METRICS = "inventory/deprecated/feeds/%s/resources/%s/metrics";
-        public static final String FEED_RESOURCES = "inventory/deprecated/feeds/%s/resources";
-        public static final String FEED_REC_RESOURCES = "inventory/deprecated/feeds/%s/resources/%s/children";
-        public static final String METRICS = "inventory/deprecated/%s/resources/%s/metrics";
+        public static final String ENVIRONMENTS = "inventory/traversal/type=e";
+        public static final String FEEDS = "inventory/traversal/type=f";
+        public static final String FEED_METRICS = "inventory/traversal/f;{f}/{r}/rl;incorporates/type=m";
```
metlos, line 65: This is equivalent to what you had before but I wonder about naming of this... It's not actually a feed metric you're looking for using this, but a metric of a resource in a feed.

mobile/src/main/java/org/hawkular/client/android/backend/BackendPipes.java
```diff
@@ -60,19 +60,19 @@ private Paths() {
...
+        public static final String FEED_RESOURCES = "inventory/traversal/f;{f}/type=r";
+        public static final String FEED_CHILD_RESOURCES = "inventory/traversal/f;{f}/{r}/type=r";
+        public static final String METRICS = "inventory/traversal/e;{e}/{r}/type=m";
```
metlos: `inventory/traversal/e;{e}/{r}/rl;incorporates/type=m`. Again, naming. This is not "metrics". This is metrics in a resource in some environment.

mobile/src/main/java/org/hawkular/client/android/backend/BackendPipes.java
```diff
@@ -60,19 +60,19 @@ private Paths() {
...
         public static final String METRIC_DATA_AVAILABILITY= "metrics/availability/%s/data";
         public static final String METRIC_DATA_COUNTER = "metrics/counters/%s/data";
         public static final String METRIC_DATA_GAUGE = "metrics/gauges/%s/data";
-        public static final String OPERATIONS = "inventory/deprecated/feeds/%s/resourceTypes/%s/operationTypes";
+        public static final String OPERATIONS = "inventory/traversal/f;{f}/rt;{rt}/type=ot";
         public static final String PERSONA = "hawkular/metrics/tenants";
         public static final String PERSONAS = "hawkular/metrics/tenants";
-        public static final String RESOURCES = "inventory/deprecated/%s/resources";
```
metlos, line 75: What was `%s` replaced by in your old impl? You don't have any placeholder in your new version...

openwarrior, line 75: earlier formatting of String was used but now it is using string replacements for {..?} So that everything can be handled inside CanonicalPath class of this client

mobile/src/main/java/org/hawkular/client/android/backend/BackendPipes.java
openwarrior: Yeah. Naming is in a mess. Sooner to be fixed.

mobile/src/main/java/org/hawkular/client/android/backend/BackendPipes.java
metlos, line 75: But what was `%s` replaced by? I don't actually see it used at all in this PR so that's why I ask. Also `/traversal/type=r` will give you nothing, because the tenant cannot contain the resources directly. The query `/traversal/recursive/type=r` would give you ALL resources under the tenant (including the child resources of other resources) but that is a super-heavy query that I would not recommend doing often.

metlos, line 75: If you wanted to find just the top-level resources, you could:
`/traversal/type=f;type=e/type=r` which give you all the resources directly contained by all environments and feeds. It would not return you the child resources of those resources though.

mobile/src/main/java/org/hawkular/client/android/backend/BackendClient.java
```diff
@@ -234,42 +234,40 @@ public void getFeeds(@NonNull Callback<List<Feed>> callback) {
...
-                Uris.getEncodedParameter(feed.getId())));
+        URI uri = Uris.getUri(CanonicalPath.getByString(feed.getPath()).fix(BackendPipes.Paths.FEED_RESOURCES));
 
         readPipe(BackendPipes.Names.FEED_RESOURCES, uri, callback);
     }
 
 
     public void getRecResourcesFromFeed(@NonNull Callback<List<Resource>> callback, Resource resource) {
 
-            URI uri = Uris.getUri(String.format(BackendPipes.Paths.FEED_REC_RESOURCES,resource.getFeed(),
```
openwarrior, line 254: This is what was happening earlier

openwarrior, line 254: All resources are no longer used in code anywhere. So I think it would be good to remove them. Adding commit for renaming those metrics and removing all resource collection part

pilhuhn:
Thank you @metlos for reviewing

openwarrior:
@pilhuhn removing resource related redundant code will be too much for this PR. Will send that in coming PR

PR #85, opened 2017-01-13
Chage URI used for checking credentials

openwarrior:
- change uri used for checking cred
- remove fetching persona


openwarrior:
Oops. Correcting license headers

PR #95, opened 2017-03-02
fixed #93 added supported hawkular version

shubhamvashisht:

PR #96, opened 2017-03-02
fix : show message for empty list

jainadit27:
Check for empty list


KeenWarrior:
Check what you did wrong from https://github.com/hawkular/hawkular-android-client/pull/95#issuecomment-283667995

KeenWarrior:
Issue is closed with #97

PR #97, opened 2017-03-03
Shows message when the favourite triggers list is empty

### LINKED ISSUE #92: Blank Screen when favourite triggers list is empty
Screen should show message about list being empty.

Changes needed to be done in

`src/main/java/org/hawkular/client/android/fragment/FavTriggersFragment.java`

Refer #91 for help.

sauravvishal8797:
Fixed #92 enabled showing message if the favourite triggers list is empty.

### DIFF
--- a/mobile/src/main/java/org/hawkular/client/android/fragment/FavTriggersFragment.java
+++ b/mobile/src/main/java/org/hawkular/client/android/fragment/FavTriggersFragment.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2015-2016 Red Hat, Inc. and/or its affiliates
+ * Copyright 2015-2017 Red Hat, Inc. and/or its affiliates
  * and other contributors as indicated by the @author tags.
  *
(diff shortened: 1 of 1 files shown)


mobile/src/main/java/org/hawkular/client/android/fragment/FavTriggersFragment.java
```diff
@@ -106,6 +111,10 @@ private void hideRefreshing() {
     private void showList() {
         ViewDirector.of(this).using(R.id.animator).show(R.id.content);
     }
+    private void showMessage(){
+        ViewDirector.of(this).using(R.id.animator).show(R.id.message);
+
```
KeenWarrior: Do not hate me for this but will be better you remove this blank line

sauravvishal8797: Will do it now.

PR #101, opened 2017-03-04
fix #100 : Wrong Navigation Item Checked

### LINKED ISSUE #100: Wrong Navigation Item Checked
ezgif com-resize

When I return from setting option to the DrawerActivity, I found that setting option is still checked. This issue is happening with Help & feedback and Explorer option.

sagar15795:


mobile/src/main/java/org/hawkular/client/android/activity/DrawerActivity.java
```diff
@@ -16,9 +16,6 @@
  */
 package org.hawkular.client.android.activity;
 
-import java.util.ArrayList;
```
KeenWarrior: Hello @sagar15795 why was relocation of imports needed? I feel you forgot to import settings.jar in Android studio before applying reformat.

It will be very help full if you correct this by first importing the jar as instructed in contribution section.

Pre-compiled jar is at https://github.com/garg-anuj/hawkular-android-client-extras/blob/master/ide-configs/settings.jar

sagar15795: @garg-anuj 
I have import settings.jar to the studio, but it is not working. For now, I have to remove the issue manually.

KeenWarrior:
Thanks @sagar15795

PR #102, opened 2017-03-04
fixed #98 added progressDialog while authenticating

### LINKED ISSUE #98: implement loading spinner during authentication to show progress.
when we log in to the server there is no visual hint that user is being authenticated. so user tries to click login button again and again in time it takes to log in. so there should be a loading icon or something similar to show when login /authentication is going on.

shubhamvashisht:


KeenWarrior:
Hi @shubhamvashisht  I see you usually work on your master branch. That's probably the reason you start every PR with a merge from origin/master.
Please have a look at https://guides.github.com/introduction/flow/ which describes the approach we're trying to follow (except the deploy step).

PR #108, opened 2017-03-05
fixed #106 message for empty list in alertFragment

### LINKED ISSUE #106: Message for empty list not visible after switching between include resolved and remove reslved
At https://github.com/hawkular/hawkular-android-client/blob/master/mobile/src/main/java/org/hawkular/client/android/fragment/AlertsFragment.java#L255

Apply similar as of #97 based on size of this.alerts

shubhamvashisht:


mobile/src/main/java/org/hawkular/client/android/fragment/AlertsFragment.java
```diff
@@ -434,8 +438,10 @@ public void onSuccess(List<Alert> alerts) {
         @Override
         public void onFailure(Exception e) {
             Timber.d(e, "Alerts fetching failed.");
+            if (getAlertsFragment()!=null){
```
KeenWarrior: This check will never be needed

mobile/src/main/java/org/hawkular/client/android/fragment/AlertsFragment.java
```diff
@@ -441,7 +445,7 @@ public void onFailure(Exception e) {
         private AlertsFragment getAlertsFragment() {
             return (AlertsFragment) getSupportFragment();
         }
-    }
```
KeenWarrior: Please remove this diff too

shubhamvashisht: Done @garg-anuj

PR #109, opened 2017-03-05
placed the plus action of alerts fragment in overflow and enabled titl…

### LINKED ISSUE #105: On alerts screen + action is confusing
On alerts fragment plus action denoting id `show_hide_res` for showing and hiding toggle for resolved alert should be placed in overflow where its title should change according to the state.

Titles should be `Show Resolved` and `Hide Resolved`

sauravvishal8797:
fixed #105 placed the plus action in alerts fragment in overflow and enabled title change according to state.


KeenWarrior:
Thanks @sauravvishal8797

PR #111, opened 2017-03-06
Upgrade Aerogear UPS integration and add FCM

KeenWarrior:
- Upgrade aerogear-android-push to 4.0.0
- Add FCM support

### DIFF
--- a/mobile/build.gradle
+++ b/mobile/build.gradle
@@ -27,6 +27,10 @@ android {
         buildConfigField "String", "UPS_VARIANT", formatStringField(readProperty(serverProperties, "ups.variant"))
         buildConfigField "String", "UPS_SECRET", formatStringField(readProperty(serverProperties, "ups.secret"))
         buildConfigField "String", "GCM_SENDER", formatStringField(readProperty(serverProperties, "gcm.sender"))
+        buildConfigField "String", "GCM_APPLICATION_ID", formatStringField(readProperty(serverProperties, "gcm.application"))
+        buildConfigField "String", "GCM_API_KEY", formatStringField(readProperty(serverProperties, "gcm.api"))
+        buildConfigField "String", "GCM_DATABASE_URL", formatStringField(readProperty(serverProperties, "gcm.database"))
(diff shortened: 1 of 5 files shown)


mobile/build.gradle
```diff
@@ -140,7 +144,7 @@ dependencies {
     compile "info.metadude.android:typed-preferences:2.0.0"
     compile "org.jboss.aerogear:aerogear-android-authz:3.0.0"
     compile "org.jboss.aerogear:aerogear-android-pipe:3.1.0"
-    compile "org.jboss.aerogear:aerogear-android-push:3.0.0"
+    compile "org.jboss.aerogear:aerogear-android-push:4.0.0"
```
danielpassos: Any reason to use `4.0.x` instead of `4.1.x`. If so please use the last bug fix version of `4.0.x` (`4.0.2`). 4.0.0 is buggy!

danielpassos (changes requested):

mobile/src/main/java/org/hawkular/client/android/push/PushClient.java
```diff
@@ -56,7 +59,14 @@ public void setUpPush() {
             return;
         }
 
-        RegistrarManager.config(PushConfiguration.NAME, AeroGearGCMPushConfiguration.class)
+        FirebaseApp.initializeApp(this.context, new FirebaseOptions.Builder()
```
danielpassos, line 62: Do we really need this? I'm asking because I wanna move the properties values to the json like this one https://github.com/aerogear/aerogear-android-cookbook/blob/master/HelloPush/app/src/main/assets/push-config.json

KeenWarrior: I added FirebaseApp.initializeApp so that we do not need to hard code values in app via google-services.json

With this we will later be able to fetch these data from REST endpoint and will let admins use there own UPS with firebase

mobile/build.gradle
KeenWarrior: Not confirmed if issue I faced was because of 4.1.x. Updating to 4.0.2 Will update to 4.1.x later after surety

danielpassos (approved):

PR #116, opened 2017-03-07
Create generic detail fragment class

KeenWarrior:
- Move generic code of alert detail fragment class to generic detail fragment class

### DIFF
--- a/mobile/src/main/java/org/hawkular/client/android/activity/DetailActivity.java
+++ b/mobile/src/main/java/org/hawkular/client/android/activity/DetailActivity.java
@@ -1,0 +1,72 @@
+/*
+ * Copyright 2015-2017 Red Hat, Inc. and/or its affiliates
+ * and other contributors as indicated by the @author tags.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *    http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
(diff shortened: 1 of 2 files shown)


mobile/src/main/java/org/hawkular/client/android/activity/DetailActivity.java
```diff
@@ -0,0 +1,85 @@
...
+    @Override
+    protected void onCreate(Bundle savedInstanceState) {
+        super.onCreate(savedInstanceState);
+        setContentView(R.layout.activity_container);
+
+        setUpBindings();
+
+        setUpToolbar();
+
+        setUpDetails();
```
danielpassos: I know it's a legacy, but I'd like to kill that kind of unnecessary methods, like we started to do last year

`setUpBindings`, `setUpToolbar`, `setUpDetails`

danielpassos (changes requested):

mobile/src/main/java/org/hawkular/client/android/activity/DetailActivity.java
KeenWarrior: @danielpassos Yes we will change that. Right now just preparing environment so new comers can start contributing. In this PR I just moved the code from main to generic class. Should I also cure it in this one?

danielpassos: Why not? Let fix whatever we can in new PR's. Kill those methods and move all contents to the `onCreate`

It will make me a _little_ happy ;)

KeenWarrior: @danielpassos Will do that tomorrow for sure. Right now a kid is stuck because of this. 

So can you please let it pass for a day.

Just a day ;)

KeenWarrior: Its 1 AM already. I am not feeling energetic enough to pull up my laptop

danielpassos (approved):

PR #126, opened 2017-03-09
fixed #119 Replaced GCM by FCM

### LINKED ISSUE #119: Replace GCM by FCM
There is some variable and properties using `GCM` instead of `FCM`. Since `GCM` is replaced by `FCM` we should also replace that variables and properties

shubhamvashisht:


shubhamvashisht:
@garg-anuj  @danielpassos  please review.

KeenWarrior (approved):
Good work

PR #137, opened 2017-03-10
Move license check from maven to gradle

KeenWarrior:


pilhuhn (commented):
I don't think a general "reformat" - here license year change should be in the same commit/push as functional changes.
Please split that.

On top I am not convinced that the bulk year change is really wanted/needed. But then I know others have other opinions here.

Issue #124, opened 2017-03-09
Create a toggle to represent and change state of trigger from TriggersActivity


KeenWarrior:
- This issue will need #113 to finish first
- Crate a toggle in place of initial popup area

KeenWarrior:
@shubhamvashisht You can go for it.

shubhamvashisht:
@garg-anuj  Thanks

KeenWarrior:
@shubhamvashisht Read data type `Trigger` http://www.hawkular.org/docs/rest/rest-alerts.html#Trigger
We would need create a PUT call changing `enabled` attribute when toggle is switched.

Steps would be
- Create toggle UI
- Create PUT endpoint call for method Trigger
- Create callback

shubhamvashisht:
@garg-anuj  please check

shubhamvashisht:
@garg-anuj  i need help in the feature you suggested for disabling toggles when state gets changed.

KeenWarrior:
@shubhamvashisht Thanks

PR #148, opened 2017-03-15
Create a toggle to represent and change state of trigger from TriggesActivity #124 :step-1

shubhamvashisht:
i have commented out some code which will be needed in next step.

### DIFF
--- a/mobile/src/main/java/org/hawkular/client/android/adapter/TriggersAdapter.java
+++ b/mobile/src/main/java/org/hawkular/client/android/adapter/TriggersAdapter.java
@@ -24,6 +24,7 @@ import org.hawkular.client.android.backend.model.Trigger;
 
 import android.content.Context;
 import android.support.annotation.NonNull;
+import android.support.v7.widget.SwitchCompat;
 import android.view.LayoutInflater;
 import android.view.View;
 import android.view.ViewGroup;
@@ -41,7 +42,7 @@ import butterknife.ButterKnife;
 public class TriggersAdapter extends BindableAdapter<Trigger> {
 
     public interface TriggerListener {
-        void onTriggerMenuClick(View TriggerView, int triggerPosition);
+        void onTriggerToggleChanged(View TriggerView, int triggerPosition,boolean state);
         void onTriggerTextClick(View TriggerView, int triggerPosition);
     }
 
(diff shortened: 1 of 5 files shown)


mobile/src/main/java/org/hawkular/client/android/backend/BackendPipes.java
```diff
@@ -72,6 +72,7 @@ private Paths() {
         public static final String OPERATIONS = "inventory/traversal/f;{f}/rt;{rt}/type=ot";
         public static final String PERSONA = "hawkular/metrics/tenants";
         public static final String TRIGGERS = "alerts/triggers";
+        public static final String TRIGGER_STATUS = "triggers/enabled";
```
KeenWarrior: this should be part of next PR. Please revert this from this PR

mobile/src/main/java/org/hawkular/client/android/adapter/TriggersAdapter.java
```diff
@@ -92,12 +93,13 @@ public void onClick(View view) {
             }
         });
 
-        viewHolder.menuButton.setOnClickListener(new View.OnClickListener() {
-            @Override
-            public void onClick(View view) {
-                triggerListener.onTriggerMenuClick(view, position);
+        viewHolder.toggleTrigger.setChecked(trigger.getEnableStatus());
+
+        /*viewHolder.toggleTrigger.setOnCheckedChangeListener(new CompoundButton.OnCheckedChangeListener() {
```
KeenWarrior: If this commented code is no longer needed. Remove that

KeenWarrior: I will suggest to use `onTriggerToggleChanged` only at place of `setChecked`

mobile/src/main/java/org/hawkular/client/android/backend/model/Trigger.java
```diff
@@ -74,13 +82,15 @@ public Trigger createFromParcel(Parcel parcel) {
     private Trigger(Parcel parcel) {
         this.id = parcel.readString();
         this.description = parcel.readString();
+        this.enabled = (boolean) parcel.readValue(null);
```
KeenWarrior: As from json `true` and `false will be received as Strings. You can have a look [example](https://github.com/hawkular/hawkular-android-client/blob/master/mobile/src/main/java/org/hawkular/client/android/backend/model/MetricAvailabilityBucket.java#L42) for how you should save and receive them

KeenWarrior (changes requested):

shubhamvashisht:
@garg-anuj changes done sir.

KeenWarrior (approved):
Looking good to me

PR #154, opened 2017-03-17
Add TriggerDetail model

KeenWarrior:


danielpassos:
@garg-anuj I don't like to use `Parceble` but we will discuss about that later

danielpassos (approved):

Issue #156, opened 2017-03-17
Move checkstyle from maven goal to gradle


KeenWarrior:
Use [gradle plugin](https://docs.gradle.org/current/userguide/checkstyle_plugin.html) to shift checkstyle to gradle

[checkstyle.xml](https://raw.githubusercontent.com/hawkular/hawkular-build-tools/master/build-tools/src/main/resources/hawkular-checkstyle/checkstyle.xml)

[Checkstyle settings](https://github.com/hawkular/hawkular-parent-pom/blob/master/pom.xml#L99-L139)

pg301:
@garg-anuj can I try

KeenWarrior:
@pg301 Sure

shubhamvashisht:
i would like to  work.

KeenWarrior:
@shubhamvashisht One work at a time ;) I will get something else for you when you finish this one

KeenWarrior:
@shubhamvashisht I mean your present one :smile:

shubhamvashisht:
@garg-anuj  Alright sir

sauravvishal8797:
@garg-anuj I would also like to work on it.

janakamarasena:
@garg-anuj I would also like to work on this.

danielpassos:
@sauravvishal8797 @janakamarasena Sorry, @pg301 Is already on it. I just assigneed that to her to avoid confusion

KeenWarrior:
Closed with #159

PR #159, opened 2017-03-17
Move checkstyle from maven goal to gradle #156

pg301:
Move checkstyle from maven goal to gradle #156


pg301:
WIP

mobile/build.gradle
```diff
@@ -122,6 +123,17 @@ def readProguardFiles() {
     return proguardFiles
 }
 
+tasks.withType(Checkstyle) {
+    configFile = rootProject.file('config/checkstyle/checkstyle.xml')
```
KeenWarrior: It will be better to have `checkstyle.xml` under `misc/`

mobile/build.gradle
```diff
@@ -122,6 +123,17 @@ def readProguardFiles() {
...
+
+    source 'src'
+    include '**/*.java'
```
KeenWarrior: also include `**/*.json`, `**/*.xml`

mobile/build.gradle
pg301: @garg-anuj Gradle guidelines say that checkstyle.xml should be in a folder called config. I can transfer the file if you want.

KeenWarrior: @pg301 When it is at `config/checkstyle/checkstyle.xml` then it is not needed to be specified. That is the magic of convention. But creating multiple folders for such minimal things will also be wrong way. So please move them to `misc/` :wink:

pg301:
@garg-anuj @danielpassos Some more changes required?

KeenWarrior:
@pg301 I will test and let you know

KeenWarrior (approved):

Issue #174, opened 2017-03-20
Toast while making favorite and when toggle button is clicked


anant-dev:
we can add a toast of short description to let user know what happened, on clicking the heart in individual trigger and on clicking the toggle button for respective trigger.

danielpassos:
I'd like to use `Snackbar` instead of `Toast`

anant-dev:
@danielpassos Ok i will make the changes and make a new pr

danielpassos:
@anant-dev You don't need to make a new one. You can update the current one!

shubhamvashisht:
@danielpassos sir, guided by @garg-anuj i am already on something similar to it. i will update the issue #124 .

danielpassos:
ok

anant-dev:
@danielpassos  i have updated the pr please review.

PR #176, opened 2017-03-20
Add toast for favourites and triger toggle

anant-dev:
Fixes issue #174 Please review the pull request.

### DIFF
--- a/mobile/src/main/java/org/hawkular/client/android/fragment/TriggerDetailFragment.java
+++ b/mobile/src/main/java/org/hawkular/client/android/fragment/TriggerDetailFragment.java
@@ -28,6 +28,7 @@ import org.jboss.aerogear.android.store.sql.SQLStoreConfiguration;
 import android.content.Context;
 import android.os.Bundle;
 import android.support.annotation.Nullable;
+import android.support.design.widget.Snackbar;
 import android.view.LayoutInflater;
 import android.view.View;
 import android.view.ViewGroup;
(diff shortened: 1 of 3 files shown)


danielpassos:
@anant-dev Can you please replace `Toast` by `Snackbar`

mobile/src/main/java/org/hawkular/client/android/fragment/TriggerDetailFragment.java
```diff
@@ -72,10 +73,12 @@ public void toggleTriggerFavouriteStatus(){
 
         if (store.read(trigger.getId()) == null) {
             store.save(trigger);
+            Toast.makeText(getContext(),"Added to Favourites", Toast.LENGTH_SHORT).show();
```
danielpassos: All strings should be provided from `strings.xml`

m-murad: Try not to hard code the strings.
Instead user the strings.xml resources file to save strings so that the application can be used by non english speaking people.
Just an opinion

danielpassos (changes requested):

mobile/src/main/java/org/hawkular/client/android/fragment/TriggerDetailFragment.java
anant-dev: ok

danielpassos (approved):

mobile/src/main/java/org/hawkular/client/android/fragment/TriggerDetailFragment.java
anant-dev: ok i have taken care of it in the new commit. Thanks for the suggestion :+1:

PR #183, opened 2017-03-21
Resource leak fixed.

m-murad:
Fixes resource leak in `TriggersFragment`.


KeenWarrior:
@free4murad It will be awesome if you roam around entire app and resolve all resource leaks at once

KeenWarrior (approved):

Issue #186, opened 2017-03-22
App crashes on adding a metric to favourite


pg301:
`E/AndroidRuntime: FATAL EXCEPTION: main
                  Process: org.hawkular.client.android, PID: 2860
                  java.lang.IllegalStateException: Required view 'button_menu' with ID 2131624118 for field 'menuButton' was not found. If this view is optional add '@Nullable' (fields) or '@Optional' (methods) annotation.
                      at butterknife.internal.Utils.findRequiredView(Utils.java:92)
                      at org.hawkular.client.android.adapter.FavMetricsAdapter$ViewHolder_ViewBinding.<init>(FavMetricsAdapter$ViewHolder_ViewBinding.java:24)
                      at java.lang.reflect.Constructor.newInstance0(Native Method)
                      at java.lang.reflect.Constructor.newInstance(Constructor.java:430)
                      at butterknife.ButterKnife.createBinding(ButterKnife.java:199)
                      at butterknife.ButterKnife.bind(ButterKnife.java:172)
                      at org.hawkular.client.android.adapter.FavMetricsAdapter$ViewHolder.<init>(FavMetricsAdapter.java:118)
                      at org.hawkular.client.android.adapter.FavMetricsAdapter.newView(FavMetricsAdapter.java:79)
                      at org.hawkular.client.android.adapter.BindableAdapter.getView(BindableAdapter.java:45)
                      at android.widget.AbsListView.obtainView(AbsListView.java:2363)
                      at android.widget.ListView.makeAndAddView(ListView.java:1970)
                      at android.widget.ListView.fillDown(ListView.java:704)
                      at android.widget.ListView.fillFromTop(ListView.java:765)
                      at android.widget.ListView.layoutChildren(ListView.java:1716)
                      at android.widget.AbsListView.onLayout(AbsListView.java:2162)
                      at android.view.View.layout(View.java:17637)
                      at android.view.ViewGroup.layout(ViewGroup.java:5575)
                      at android.support.v4.widget.SwipeRefreshLayout.onLayout(SwipeRefreshLayout.java:636)
                      at android.view.View.layout(View.java:17637)
                      at android.view.ViewGroup.layout(ViewGroup.java:5575)
                      at android.widget.FrameLayout.layoutChildren(FrameLayout.java:323)
                      at android.widget.FrameLayout.onLayout(FrameLayout.java:261)
                      at android.view.View.layout(View.java:17637)
[...]

PR #187, opened 2017-03-22
fixed app crash on adding a metric to favourite #186

pg301:


mobile/src/main/java/org/hawkular/client/android/adapter/FavMetricsAdapter.java
```diff
@@ -75,7 +75,7 @@ public long getItemId(int position) {
     @NonNull
     @Override
     protected View newView(LayoutInflater inflater, ViewGroup viewContainer) {
-        View view = inflater.inflate(R.layout.layout_list_item_token, viewContainer, false);
+        View view = inflater.inflate(R.layout.layout_list_item_token_with_menu, viewContainer, false);
```
KeenWarrior: Can you please also refactor file with name layout_list_item_with_menu.

KeenWarrior (changes requested):

mobile/src/main/res/layout/layout_list_item_token_with_menu.xml
```diff
@@ -0,0 +1,67 @@
...
+
+    <ImageView
+         android:id="@+id/button_menu"
+         app:srcCompat="@drawable/ic_more"
+         android:scaleType="center"
+         android:background="?attr/selectableItemBackgroundBorderless"
+         android:layout_width="48dp"
+         android:layout_height="48dp"/>
+
+</LinearLayout>
```
anant-dev: @garg-anuj @pg301 no need to create a new layout, previous layout can be easily reused. Please view issue #186 comments

pg301: @anant-dev I thought about that but then we would have to change the visibility depending on the fragment called. For eg if we are in FavMetricFragment then we don't want toggle , so we would have to set the visibility = gone for toggleswitch. I thought unnecessary calling of UI methods might be heavy on thread so I created a new list item.

anant-dev: @pg301 so we can use layout_list_item_alert for the purpose because the only difference between your newly created file and layout_list_item_alert is the id of linear layout which in the file case is data_box and your case is list_item which can easily be changed in  FavMetricFragment

anant-dev: resolved this in 5ed52795aef3f8092171f2e214dc261305b80a7f commit pls check @pg301 @garg-anuj

pg301: @anant-dev I didn't check the list_item_alert. My bad

KeenWarrior:
@pg301 You can borrow the work @anant-dev and complete this PR. 

- Refactor list_item_alert into list_item_with_menu
- Provide it in FavMetricsAdapter.java

KeenWarrior:
@pg301 you will need to rebase

KeenWarrior:
@pg301 Rebasing this branch will need more effort than work in PR. You can either reset this branch with current master and work on that to resolve this issue. Or altogether create a new branch of master

pg301:
@garg-anuj Is it ok if I make a new new PR for this issue and close the current PR.

KeenWarrior:
@pg301 Sure

pg301:
I am closing this PR since I have made a new PR addressing this issue.

Issue #205, opened 2017-04-14
TODOs in the code base.


m-murad:
While going through the source code, I noticed there are certain `TODOs` marked in the project. 
@pilhuhn @garg-anuj @danielpassos 
Can we discuss these.
![screen shot 2017-04-14 at 11 49 41 am](https://cloud.githubusercontent.com/assets/17262180/25035025/80a7f502-2108-11e7-9d07-216f5d8c3591.png)

I would like to send a PR for this.

pg301:
@free4murad good idea.  We both can work on it, once we discuss these tasks with @danielpassos @garg-anuj @pilhuhn

shubhamvashisht:
@pg301 @free4murad   i would like to help implementing some tasks too.  @danielpassos @garg-anuj @pilhuhn

KeenWarrior:
we can discuss on gitter

PR #208, opened 2017-04-14
View unbind() added.

m-murad:
Fixes part of #205 

please review

### DIFF
--- a/mobile/src/main/java/org/hawkular/client/android/fragment/AlertsFragment.java
+++ b/mobile/src/main/java/org/hawkular/client/android/fragment/AlertsFragment.java
@@ -56,6 +56,7 @@ import android.view.ViewGroup;
 import butterknife.BindView;
 import butterknife.ButterKnife;
 import butterknife.OnClick;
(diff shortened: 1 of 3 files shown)


mobile/src/main/java/org/hawkular/client/android/fragment/AlertsFragment.java
```diff
@@ -52,9 +32,31 @@
 import android.view.MenuItem;
 import android.view.View;
 import android.view.ViewGroup;
+
+import org.hawkular.client.android.R;
```
KeenWarrior: Reordering and reformating should be avoided with other changes

m-murad: Actually to avoid unused imports I optimised the imports.
That's how this reformation must have happened.
Will take care from now onwards

m-murad:
@garg-anuj 
I have updated the PR. 
Please check.

KeenWarrior (approved):

Issue #210, opened 2017-04-16
Multiple instances of a Fragment.


m-murad:
If we select a Fragment form the NavigationBar which is already open a new instance of the fragment is created. This causes unnecessary extra use of memory.

This needs to be fixed.
I'm sending a PR for this.

m-murad:
@garg-anuj 
This has been fixed, closing this.

PR #211, opened 2017-04-16
Fixed multiple instances of a single Fragment.

m-murad:
Fixes issue #210 
Please review.

### DIFF
--- a/mobile/src/main/java/org/hawkular/client/android/activity/MainActivity.java
+++ b/mobile/src/main/java/org/hawkular/client/android/activity/MainActivity.java
@@ -299,6 +299,9 @@ public final class MainActivity extends AppCompatActivity
     }
 
     private void showFavourites() {
+        if (atHome) {
+            return;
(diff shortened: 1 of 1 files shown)


KeenWarrior:
@free4murad Can you clearly address what was the issue before

m-murad:
@garg-anuj 
If in the navbar the option selected is `Aletrs ` and the user selects  `Aletrs` again a new instance of AlertFragment and TriggersFragment is initialised.
This behavior can be seen using the memory monitor in Android studio.

What should happen is, if a fragment is in memory it should not be reinstentiated.

m-murad:
@garg-anuj 
You can see this behavior in many apps eg Gmail.

If you are in the Primary Inbox and select Primary Inbox from navbar the page is not reloaded. If you select some other option you will observe a small load time. But when selecting same option there is no load time.

KeenWarrior (approved):

PR #223, opened 2017-06-19
Added Floating Action Button for Alert Setup

pg301:
Added new fragment to display the setup form
Added some setters in Trigger model class
Bindings setup and data extraction from the form

### DIFF
--- a/mobile/src/main/java/org/hawkular/client/android/activity/MainActivity.java
+++ b/mobile/src/main/java/org/hawkular/client/android/activity/MainActivity.java
@@ -42,7 +42,6 @@ import org.hawkular.client.android.backend.BackendClient;
 import org.hawkular.client.android.backend.model.Persona;
 import org.hawkular.client.android.event.Events;
 import org.hawkular.client.android.explorer.InventoryExplorerActivity;
-import org.hawkular.client.android.fragment.AlertSetupFragment;
 import org.hawkular.client.android.fragment.FavMetricsFragment;
 import org.hawkular.client.android.fragment.TriggersFragment;
 import org.hawkular.client.android.util.Fragments;
@@ -305,13 +304,7 @@ public final class MainActivity extends AppCompatActivity
     @OnClick(R.id.fab)
     void onClickFab(){
 
-        Bundle bundle = new Bundle();
-        bundle.putString("state", "From Home");
-        AlertSetupFragment alertSetupFragment = new AlertSetupFragment();
-        alertSetupFragment.setArguments(bundle);
-
-        adapter.addFragment(alertSetupFragment,"Setup Alert");
-        adapter.notifyDataSetChanged();
+        startActivity(new Intent(getApplicationContext(), TriggerSetupActivity.class));
         floatingActionButton.setVisibility(View.GONE);
     }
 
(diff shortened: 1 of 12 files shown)


pg301:
screenshot_1497853318

On clicking FAB, I am adding a new tab in the main activity viewpager with the setup alert fragment.

mobile/src/main/java/org/hawkular/client/android/activity/MainActivity.java
```diff
@@ -294,6 +300,21 @@ void openDrawerMenu() {
...
+
+    @OnClick(R.id.fab)
+    void onClickFab(){
+
+        Bundle bundle = new Bundle();
+        bundle.putString("state", "From Home");
+        AlertSetupFragment alertSetupFragment = new AlertSetupFragment();
+        alertSetupFragment.setArguments(bundle);
+
+        adapter.addFragment(alertSetupFragment,"Setup Alert");
```
pilhuhn: I think this should be "Setup Trigger".
Shouldn't that be externalised to strings.xml though?

pilhuhn:
@jshaughn Can you please have a look at the above list of parameters and help @pg301 with what makes sense ? I could imagine for example that the ID is created behind the scenes

pg301:
@pilhuhn 

I followed this documentation while making the form. The ids are generated automatically if not explicitly defined by the user.  If there are changes required then please let me know. 
screen shot 2017-06-19 at 10 05 58 pm

mobile/src/main/java/org/hawkular/client/android/activity/MainActivity.java
pg301: As you can see in the screenshot uploaded above the setup trigger is a fragment inside a tabbed viewpager. Actually, the current main activity layout has support for only tabs with viewpager so I had to add a new tab to setup trigger. This is temporary since I and @garg-anuj are still not sure about the placement of the fragment.

pg301:
@pilhuhn I wanted to give the user total control over editing the trigger values, so I have included all possible editable options in the form. I will remove some in case they aren't needed.

jshaughn:
'Setup Trigger' is probably best.  If preferred the 'Id' field can be dropped and if not supplied to the REST call it will be assigned a UUID.  If this is a UI for defining triggers it will need a way to define a condition.  We should probably start simple and perhaps allow only a single-condition trigger.  As soon as we allow multiple conditions we'd need the 'Matching' option: ALL | ANY.  We could eliminate AutoResolve for now because AutoResolve itself requires its own condition(s) and has a few options.  It may be best to have that option open up a dedicated section.

I didn't know we had any sort of alerting UI being worked on, so that's pretty cool in general.

KeenWarrior:
@pg301 This is not looking nice. Lets move creating trigger to child activity. And fab to add these should preferably be not on fav activity but on activity fetching from remote.

pg301:
@pilhuhn @jshaughn I will drop the id field and it will be a single condition trigger for now. @garg-anuj will make it a child activity.

KeenWarrior:
@pg301 Add toolbar with describing  activity and back arrow to parent activity

pg301:
@garg-anuj Sure!

pg301:
changes


Here's a GIF showing the requested changes

KeenWarrior (approved):
:+1: thanks @pg301

PR #228, opened 2017-07-04
Shifting to Retrofit along with TriggerSetup functionality

pg301:
@garg-anuj Please have a look at the PR.

### DIFF
--- a/mobile/src/main/java/org/hawkular/client/android/backend/BackendClient.java
+++ b/mobile/src/main/java/org/hawkular/client/android/backend/BackendClient.java
@@ -101,7 +101,7 @@ public final class BackendClient {
         parameters.put(BackendPipes.Parameters.FINISH_TIME, String.valueOf(finishTime.getTime()));
 
         AlertService service = retrofit.create(AlertService.class);
-        Call call = service.get(parameters);
+        Call call = service.get();
         call.enqueue(callback);
 
     }
@@ -119,7 +119,7 @@ public final class BackendClient {
     public void acknowledgeAlert(@NonNull Alert alert,
                                  @NonNull retrofit2.Callback<List<String>> callback) {
         AlertService service = retrofit.create(AlertService.class);
-        Call call = service.postAckAlert();
+        Call call = service.AckAlert();
         call.enqueue(callback);
 
     }
@@ -138,13 +138,13 @@ public final class BackendClient {
 
     public void updateTrigger(@NonNull Trigger trigger, @NonNull retrofit2.Callback<List<String>> callback){
         TriggerService service = retrofit.create(TriggerService.class);
-        Call call = service.postUpdateTrigger(trigger);
+        Call call = service.updateTrigger(trigger);
         call.enqueue(callback);
     }
 
     public void createTrigger(@NonNull FullTrigger trigger, @NonNull retrofit2.Callback<List<String>> callback){
         TriggerService service = retrofit.create(TriggerService.class);
-        Call call = service.postCreateTrigger(trigger);
+        Call call = service.createTrigger(trigger);
         call.enqueue(callback);
     }
 
@@ -166,7 +166,7 @@ public final class BackendClient {
     public void getResourcesFromFeed(@NonNull retrofit2.Callback<List<Resource>> callback, @NonNull InventoryResponseBody body){
 
         TriggerService service = retrofit.create(TriggerService.class);
-        Call call = service.postGetResourcesFromFeed(body);
+        Call call = service.getResourcesFromFeed(body);
         call.enqueue(callback);
     }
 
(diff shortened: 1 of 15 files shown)


mobile/src/main/java/org/hawkular/client/android/backend/BackendClient.java
```diff
@@ -201,9 +210,15 @@ public void getMetricData(@NonNull Metric metric, long bucket,
 
     public void getTriggers(@NonNull Callback<List<Trigger>> callback) {
         TriggerService service = retrofit.create(TriggerService.class);
-        Call call = service.get();
+        Map<String, String> map = new HashMap<>();
```
KeenWarrior: These already have been setup for application and we will need to set them for every call

mobile/src/main/java/org/hawkular/client/android/service/TriggerService.java
```diff
@@ -17,25 +17,43 @@
...
-    @PUT("hawkular/alerts/triggers/{id}")
-    Call<List<String>> updateTrigger(@Path("id") String id, @Body Trigger trigger);
-
-    @POST("hawkular/alerts/triggers/")
-    Call<List<String>> createTrigger(@Body Trigger trigger);
+    Call<List<Trigger>> get(
+            @QueryMap Map<String, String> parameters
+    );
+
+    @POST("/hawkular/alerts/triggers")
```
KeenWarrior: Maybe you are taking naming convention wrong. eg. We don't have to put post in from of method name in  `postUpdateTrigger` just because it use post http call. `updateTrigger` will look better  name. 

Secondly I doubt we will use POST call to update content. We use PUT for that.

mobile/src/main/java/org/hawkular/client/android/backend/model/Feed.java
```diff
@@ -16,67 +16,62 @@
...
  */
-public class Feed implements Parcelable {
-
-    @RecordId
-    @SerializedName("id")
-    private String id;
-
-    @SerializedName("path")
-    private String path;
+public class Feed implements Parcelable
```
KeenWarrior: Please move `{` back to line starting of block

mobile/src/main/java/org/hawkular/client/android/service/TriggerService.java
pg301: @garg-anuj I will change the names then.

mobile/src/main/java/org/hawkular/client/android/backend/model/Feed.java
pg301: Sure

mobile/src/main/java/org/hawkular/client/android/service/TriggerService.java
pilhuhn: @garg-anuj POST vs PUT is defined by the Rest-api of alerts. One has to look it up here: http://www.hawkular.org/docs/rest/rest-alerts.html#__triggers_triggers_definitions_handling

mobile/src/main/java/org/hawkular/client/android/service/AlertService.java
```diff
@@ -34,8 +34,8 @@
     Call<List<Alert>> get(@QueryMap Map<String, String> parameters);
 
     @FormUrlEncoded
-    @POST("/hawkular/alerts/ack/{alertId}")
```
KeenWarrior: This one is correct api for call for ack alert as below one do not provide {alertId}

major difference between aerogear pipe and retrofit in case of POST and PUT is, in case of pipe we need not to put the id of item we are sending on POST as it was resolved from recordid of item we are sending. But in case of retrofit we need to configure that by ourself

mobile/src/main/java/org/hawkular/client/android/service/TriggerService.java
KeenWarrior: @pilhuhn thanks. Looking at them

mobile/src/main/java/org/hawkular/client/android/service/AlertService.java
```diff
@@ -31,11 +31,11 @@
...
 
     @GET("hawkular/alerts")
-    Call<List<Alert>> get(@QueryMap Map<String, String> parameters);
+    Call<List<Alert>> get();
 
     @FormUrlEncoded
-    @POST("/hawkular/alerts/ack/{alertId}")
-    Call<List<String>> ackAlert(@Path("alertId") String alertId);
+    @POST("/hawkular/alerts/ack")
+    Call<List<String>> AckAlert();
```
KeenWarrior: @pg301 Please resolve it

pg301:
@garg-anuj Check it now.

KeenWarrior (approved):

KeenWarrior:
@pg301 thanks. Keep up the good work

PR #235, opened 2017-07-11
Remove aerogear auth

KeenWarrior:


KeenWarrior:
@shubhamvashisht @pg301 @m-murad Can you check if it is working fine?

pg301 (approved):

PR #239, opened 2017-07-19
Inventory Fixing part 4

pg301:
@garg-anuj I need your help in debugging. When I try getting data for any type of metric, it shows me that the metric.getValue returns null. Please can you check this.

### DIFF
--- a/mobile/build.gradle
+++ b/mobile/build.gradle
@@ -183,9 +183,10 @@ dependencies {
     compile "org.jboss.aerogear:aerogear-android-push:4.0.2"
     compile "com.squareup:otto:1.3.8"
     compile "frankiesardo:icepick:3.1.0"
-    compile "com.google.code.gson:gson:2.6.1"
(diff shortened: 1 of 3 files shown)


KeenWarrior:
@pg301 looking into it

pg301:
@garg-anuj I think I debugged the error. I wasn't timestamping my start and end values while making the curl call and that was giving me empty buckets.

pg301:
@pilhuhn We need to update our API documentation. http://www.hawkular.org/docs/rest/rest-metrics.html#GET__availability__id__stats mentions that this API return array of **data points**, but we get this as response https://paste.ubuntu.com/25151988/

pg301:
@garg-anuj view value is not been written for some reason.  https://github.com/pg301/hawkular-android-client/blob/6e2ee34d7c920a268969749d7dc4743fb18b7fd2/mobile/src/main/java/org/hawkular/client/android/fragment/MetricAvailabilityFragment.java#L279 

This value is always null.

KeenWarrior:
@pg301 progress in it is stuck?

PR #249, opened 2017-08-17
Alert display fixed

pg301:
screenshot_1502964907

### DIFF
--- a/mobile/src/main/java/org/hawkular/client/android/backend/model/Alert.java
+++ b/mobile/src/main/java/org/hawkular/client/android/backend/model/Alert.java
@@ -40,10 +40,10 @@ public final class Alert implements Parcelable {
     private String status;
 
     @SerializedName("ctime")
-    private long timestamp;
+    private long ctime;
 
     @SerializedName("evalSets")
-    private List<Lister> evaluations;
+    private List<List<AlertEvaluation>> evalSets;
 
     @SerializedName("notes")
     private List<Note> notes;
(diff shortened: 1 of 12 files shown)


mobile/src/main/java/org/hawkular/client/android/backend/model/Alert.java
```diff
@@ -40,10 +40,10 @@
     private String status;
 
     @SerializedName("ctime")
-    private long timestamp;
+    private long ctime;
```
pilhuhn, line 43: what about making the name real explicit? `creation_time` I know that ctime is a common abbreviation of it

pilhuhn:
lgtm

pilhuhn (approved):
Lgtm

mobile/src/main/java/org/hawkular/client/android/backend/model/Alert.java
pg301: @pilhuhn Unfortunately I cannot change the naming convention because the JSON I get has "ctime" as a parameter. Retrofit will not parse it if we give it any other name, that is the reason I changed 'timestamp' -> 'ctime'

pg301:
triggersetup

Generating Alerts

alert

pg301:
@pilhuhn @garg-anuj Please review these major changes.

KeenWarrior:
@pg301 sorry for delay :wink:

